PR #18, opened 2025-03-09
Add Github actions CI workflow & Fix testing setup

gabriel-trigo:
This pull request includes updates to dependencies and improvements to the test setup for the `smart_control` project.

Dependency updates:
* Added `pytest` to the dependencies in `pyproject.toml` to ensure proper testing framework is included.
* This workflow is set to run on pull requests to `copybara_push` branch (which we are using as the main branch)

Fixing failing tests:
* Added import of `flags` from `absl` and initialized `flags.FLAGS` in `building_utils_test.py` to avoid errors with `create_tempdir`. This fixes all the test cases that were failing

- [x] Tests pass


pyproject.toml
```diff
@@ -27,6 +27,7 @@ scikit-learn = "^1.5.1"
 ipykernel = "^6.29.5"
 typing-extensions = "^4.12.2"
 ipython = "^8.27.0"
+pytest = "^8.3.5"
```
s2t2: Since pytest is used with testing and not in the module code itself, let's prefer to add it to the development dependencies group.

Let's update the toml file using these commands:

```sh
poetry remove pytest 
poetry add --dev pytest
```

Then in the README let's update the installation command number 4 to be `poetry install --with dev`, which will install dependencies from the dev group as well.

.github/workflows/setup-test.yaml
```diff
@@ -0,0 +1,43 @@
...
+        sudo apt update
+        sudo apt install -y protobuf-compiler ffmpeg  # Install both protobuf and ffmpeg in one step
+
+    - name: Install Python dependencies
+      run: |
+        python -m pip install --upgrade pip
+        python -m venv .venv
+        source .venv/bin/activate
+        pip install poetry
+        poetry install
```
s2t2: Let's also update this to `poetry install --with dev`.

s2t2 (changes requested):
See comments about moving pytest to the dev group. Other than that this looks good.

s2t2 (approved):
Looks good. Team effort 😸

PR #19, opened 2025-03-09
feat: add reinforcement learning directory

gabriel-trigo:
The purpose of this directory is to have all reinforcement learning related code, and scripts to train and evaluate agents

To add new agents, create a new file on the `agents` directory (and any necessary networks in the `agents/networks` directory)

To train agents, use the scripts in the `scripts` directory, which allow you to populate a starter replay buffer with a policy (currently the schedule policy in `policies/schedule_policy.py`) with the `populate_starter_buffer.py` script, and train your agent with the `train.py` script

~~Before running the scripts, you must edit the `utils/config.py` file, setting the `ROOT_DIR` variable to be your directory's root path~~

- [x] Tests pass


s2t2:
It might be cleaner if you rebase this branch first. Do you have multiple git remote addresses set up? 

```
git remote add upstream git@github.com:google/sbsim.git
```

If you have your repo as the "origin" address, and the Google owned repo as the "upstream" address, you can rebase this way:

```
# updating your local repo (main branch):
git checkout copybara_push
git fetch upstream
git pull upstream copybara_push

# updating your remote repo (main branch):
git push origin copybara_push

# applying the changes to this branch:
git checkout reinforcement_learning-gabriel
git rebase -i upstream/copybara_push

# pushing to update the PR:
git push origin reinforcement_learning-gabriel
```

I can do this on my PR, but if you rebase first, our diffs will be cleaner.

smart_control/reinforcement_learning/utils/config.py
```diff
@@ -0,0 +1,119 @@
...
+from smart_control.simulator.tf_simulator import TFSimulator
+from smart_control.simulator.weather_controller import ReplayWeatherController
+from smart_control.utils import controller_reader, histogram_reducer
+from smart_control.utils.controller_writer import ProtoWriterFactory
+from smart_control.utils.environment_utils import to_timestamp
+from smart_control.utils.observation_normalizer import (
+    StandardScoreObservationNormalizer,
+)
+
+ROOT_DIR = "/home/gabriel-user/projects/sbsim/" # Path to the root directory of the project
```
s2t2: I was originally thinking about using an environment variable here to set the ROOT_DIR, but I will try a relative directory reference instead, which would simplify configuration setup and work automatically.

smart_control/reinforcement_learning/scripts/populate_starter_buffer.py
```diff
@@ -0,0 +1,180 @@
...
+    import argparse
+    
+    parser = argparse.ArgumentParser(description='Populate a replay buffer with initial exploration data')
+    parser.add_argument('--buffer-name', type=str, required=True, help='Name to identify the saved replay buffer')
+    parser.add_argument('--capacity', type=int, default=50000, help='Replay buffer capacity')
+    parser.add_argument('--steps-per-run', type=int, default=100, help='Number of steps per actor run')
+    parser.add_argument('--num-runs', type=int, default=5, help='Number of actor runs to perform')
+    parser.add_argument('--sequence-length', type=int, default=2, help='Sequence length for the replay buffer')
+    parser.add_argument('--env-gin-config-file-path', type=str, 
+                        default='/home/gabriel-user/projects/sbsim/smart_control/configs/resources/sb1/train_sim_configs/sim_config_1_day.gin',
```
s2t2: let's use the standard `CONFIG_PATH` here

smart_control/reinforcement_learning/scripts/populate_starter_buffer.py
```diff
@@ -0,0 +1,180 @@
...
+        sequence_length: Length of sequences to store in the replay buffer
+        env_gin_config_file_path: Path to the environment configuration file
+    """
+    
+    buffer_path = os.path.join(OUTPUT_DATA_PATH, f'{buffer_name}_seqlen{sequence_length}_exp{num_runs*steps_per_run}')
+    logger.info("Buffer path: %s", buffer_path)
+    
+    # Create directory if it doesn't exist
+    try:
+        os.makedirs(os.path.dirname(buffer_path + '/anything-here'), exist_ok=False) # added '/anything-here' such that
```
s2t2, line 67: is there a reason why we don't want to use the `exist_ok=True` parameter here?

s2t2: oh to avoid overriding

s2t2, line 67: @gabriel-trigo can you please further explain the "/anything-here" part of line 67?

I see the comment says we want this to be a directory. Would the following approach be sufficient? I would think the `makedirs` function would create a directory (not a file).

```python
os.makedirs(buffer_path, exist_ok=False) 
```

smart_control/reinforcement_learning/policies/schedule_policy.py
```diff
@@ -0,0 +1,256 @@
...
+        ),
+        ScheduleEvent(
+            pd.Timedelta(19, unit='hour'),
+            DeviceType.HWS,
+            'supply_water_setpoint',
+            315.0,
+        ),
+    ]
+    
+    local_start_time = env.current_simulation_timestamp.tz_convert(tz='US/Pacific')
```
s2t2: let's consider using the `DEFAULT_TIME_ZONE` here

smart_control/reinforcement_learning/observers/print_status_observer.py
```diff
@@ -0,0 +1,77 @@
...
+    This observer prints information about the training progress, including
+    rewards, execution time, and replay buffer size.
+    """
+    
+    def __init__(
+        self,
+        status_interval_steps: int=1,
+        environment=None,
+        replay_buffer=None,
+        time_zone='US/Pacific'
```
s2t2: let's consider using the `DEFAULT_TIME_ZONE` here as well

smart_control/reinforcement_learning/observers/rendering_observer.py
```diff
@@ -0,0 +1,536 @@
...
+
+class RenderingObserver(Observer):
+    """Observer that renders the environment and plots metrics.
+    
+    This observer renders the environment at specified intervals and can
+    also show plots of metrics.
+    """
+    
+    # Class constant
+    KELVIN_TO_CELSIUS = 273.15
```
s2t2: Let's consider importing this from the constants file

smart_control/reinforcement_learning/utils/data_processing.py
```diff
@@ -0,0 +1,335 @@
...
+    return pd.DataFrame({
+        'timestamp': timestamps,
+        'device_id': device_ids,
+        'setpoint_name': setpoint_names,
+        'setpoint_value': setpoint_values,
+        'response_type': response_types,
+    })
+
+
+def convert_kelvin_to_celsius(temperature_kelvin: Union[float, np.ndarray, pd.Series]) -> Union[float, np.ndarray, pd.Series]:
```
s2t2, line 314: We have some existing F to K temperature conversion functions in `smart_control.utils.conversion_utils`. Perhaps we should move these C to K conversion functions there as well?

However for now if we want to keep focus within the `smart_control.reinforcement_learning` module, that could be ok.

smart_control/reinforcement_learning/observers/base_observer.py
```diff
@@ -0,0 +1,40 @@
...
+        pass
+    
+    def reset(self) -> None:
+        """Reset the observer to its initial state.
+        
+        This method is called when a new episode starts.
+        """
+        pass
+    
+    def close(self) -> None:
```
s2t2: do we want to mark the `reset` and `close` methods as `@abc.abstractmethod` as well?

smart_control/reinforcement_learning/observers/print_status_observer.py
```diff
@@ -0,0 +1,77 @@
...
+
+import pandas as pd
+from tf_agents.trajectories import trajectory as trajectory_lib
+
+from smart_control.reinforcement_learning.observers.base_observer import Observer
+
+logger = logging.getLogger(__name__)
+
+
+class PrintStatusObserver(Observer):
```
s2t2, line 13: Does this class need a `close` method?

smart_control/reinforcement_learning/observers/rendering_observer.py
```diff
@@ -0,0 +1,536 @@
...
+    get_outside_air_temperature_timeseries,
+    get_reward_timeseries,
+    get_zone_timeseries,
+)
+from smart_control.utils import building_renderer
+
+logger = logging.getLogger(__name__)
+
+
+class RenderingObserver(Observer):
```
s2t2, line 37: Does this class need a `close` method?

smart_control/reinforcement_learning/scripts/populate_starter_buffer.py
gabriel-trigo: Yeah you are totally right here, I'm not sure this made more sense for me at the time hahaha

smart_control/reinforcement_learning/utils/data_processing.py
gabriel-trigo: Yeah I think having two different functions for the same purpose would be unnecessarily redundant. I'll look into the function you mentioned and see if we can use it easily

smart_control/reinforcement_learning/observers/base_observer.py
gabriel-trigo: That would probably be better practice. As is the subclasses can choose to implement it or not, which is not the best I guess

Makes sense to me!

smart_control/reinforcement_learning/observers/print_status_observer.py
gabriel-trigo: It should have, I forgot about it. The change in the interface to make the close and reset methods abstractmethods should prevent this since it will force those to be implemented :)

smart_control/reinforcement_learning/observers/rendering_observer.py
gabriel-trigo: It should have, I forgot about it. The change in the interface to make the close and reset methods abstractmethods should prevent this since it will force those to be implemented :)

smart_control/reinforcement_learning/utils/data_processing.py
s2t2: I don't believe there are duplicate Celsius to Kelvin functions - the existing conversion functions are Fahrenheit related. My comment is about potentially organizing all the temperature related functions together.

At this point however I would be inclined to keep all the reinforcement learning module code together, leaving these C to K functions where they are. And we can create a low priority issue to consider moving them. Then in the future if other modules need the C to K functions, that will motivate us to actually move them up into the more general `smart_control.utils.conversion_utils` location.

smart_control/reinforcement_learning/observers/base_observer.py
s2t2: @gabriel-trigo quick question - what does the close method code need to be? Is it the same for all the observers (in which case we can put it in the base class), or would it differ between observers?

I don't see the close method implemented in any of the observer classes, except for the composite observer, which calls the close method on its sub-observers (which at this point I don't believe would do anything but `pass`, as currently implemented in the base observer).

If we don't need the close method, we can remove it. But just wanted to check in here to see if you had any more information or context to share about what we would want to be doing there.

Good news is I believe this is the last thing we need to address before we merge this PR!

s2t2: Potentially related... looks like in the original notebooks, the close method was called on this observer:

```py
rb_observer = reverb_utils.ReverbAddTrajectoryObserver(
    reverb_replay.py_client, table_name, sequence_length=2, stride_length=1
)
```

In the new reinforcement learning module, it looks like a `ReverbAddTrajectoryObserver` instance has been implemented in a few places in the `ReplayBufferManager` class. But not sure how to connect the dots back to our custom observer classes from there.

gabriel-trigo: The thought here is that some observers that might have some resources that should be terminated/destroyed properly when the observer is done with its job.

The custom observers that we have don't need that, so they don't have implementations of the current method.

To be honest, I'm not completely sure how necessary this method would be, just added it because the reverb observer had so I thought it would be proper. But like you noted, the reverb observer does not implement this interface (its provided by reverb)

s2t2: OK I have implemented these changes in a follow-up PR: https://github.com/gabriel-trigo/sbsim/pull/4

Ready for your review

smart_control/reinforcement_learning/utils/data_processing.py
s2t2: #25

PR #21, opened 2025-03-22
fix: remove ambiguous @gin_configurables from SAC and TD3 notebooks

### LINKED ISSUE #13: Ambiguous Selector Error in sim_config.gin
Encountering a KeyError due to an ambiguous selector for to_timestamp, local_time and other functions defined in the __main__ in the configuration file:

Example:
KeyError: "Ambiguous selector 'to_timestamp', matches ['__main__.to_timestamp', 'smart_control.utils.environment_utils.to_timestamp']."

Which functions (from __main__ or smart_control.utils.environment_utils) should be used so I can change the sim_config.gin accordingly?

Removing the ones in the __main__ seems to be a quick fix but is that the right way it should perform?

gabriel-trigo:
Fixes #13 

Remove ambiguous `@gin_configurables` from SAC and TD3 notebooks. Now notebooks should run with no issues


smart_control/notebooks/SAC_Demo.ipynb
```diff
@@ -151,10 +151,10 @@
         "  logging.info(*args)\n",
         "  print(*args)\n",
         "\n",
-        "data_path = \"third_party/py/smart_buildings/smart_control/configs/resources/sb1/\" #@param {type:\"string\"}\n",
-        "metrics_path = \"/usr/local/google/home/metrics\" #@param {type:\"string\"}\n",
-        "output_data_path = '/usr/local/google/home/smart_buildings/smart_control/sb_colab_demo' #@param {type:\"string\"}\n",
-        "root_dir = \"/usr/local/google/root\" #@param {type:\"string\"}\n",
+        "data_path = \"/home/gabriel-user/projects/sbsim-2/sbsim/smart_control/configs/resources/sb1/\" #@param {type:\"string\"}\n",
```
s2t2, line 154: I am going to implement environment variables to help us customize these paths without updating the source code. #23

smart_control/notebooks/SAC_Demo.ipynb
```diff
@@ -1658,44 +1534,12 @@
     },
     {
       "cell_type": "code",
-      "execution_count": 12,
+      "execution_count": null,
       "metadata": {
         "cellView": "form",
         "id": "3Zv-lSiWDp50"
       },
-      "outputs": [
```
s2t2, line 1666: I wonder if we should be saving the notebook without outputs (to help with managing the code, in an effort to reduce the diffs).

smart_control/notebooks/SAC_Demo.ipynb
```diff
@@ -991,137 +991,13 @@
...
-            "  np.array(weekday_energy_prices)\n",
-            "/home/trigo/sbsim/sbsim/smart_control/reward/electricity_energy_cost.py:159: UnitStrippedWarning: The unit of the quantity is stripped when downcasting to ndarray.\n",
-            "  np.array(weekend_energy_prices)\n"
-          ]
-        }
-      ],
+      "outputs": [],
       "source": [
-        "@gin.configurable\n",
-        "def to_timestamp(date_str: str) -> pd.Timestamp:\n",
```
s2t2, line 1057: Confirm I am running into the same issue about ambiguous functions (`to_timestamp`, etc). I'm going to replicate removing these function definitions (or maybe just commenting them out) in #23

smart_control/notebooks/SAC_Demo.ipynb
```diff
@@ -65,7 +65,7 @@
         "import enum\n",
         "import functools\n",
         "import os\n",
-        "import os\n",
+        "os.environ['WRAPT_DISABLE_EXTENSIONS'] = 'true'\n",
```
s2t2, line 68: I will be enabling setting of environment variables via the .env file in #23 so each person can toggle this at runtime without modifying the source code.

I am a little unfamiliar with this option, so not sure if we would always want it set to true. But I read there could be performance gains with the C extensions enabled, so I'm inclined to leave it configurable (via the .env file). 

If anyone has any thoughts or context to share about this option, let me know.


Info from Gemini:

> The wrapt library is a Python module used for creating decorators and wrappers. It often uses C extensions for performance reasons. These C extensions can provide significant speed improvements in certain scenarios.
>
> However, there might be situations where you want to disable these C extensions. Setting the WRAPT_DISABLE_EXTENSIONS environment variable to 'true' tells the wrapt library to fall back to its pure Python implementation instead of using the faster C extensions.

gabriel-trigo: This was added a while ago because it turns out something breaks unless you set it like this. I forgot what it was exactly, but if you try running the notebook without setting this, I am pretty confident you will run into an error

s2t2: OK cool, thanks. We can leave this set then.

smart_control/notebooks/SAC_Demo.ipynb
s2t2: For now we can keep these diffs. I want to prioritize merging this PR first and then I will implement environment variables afterwards.

smart_control/notebooks/SAC_Demo.ipynb
s2t2: We have decided we do want to save notebooks without outputs. If we need to showcase results we can create a PDF or something.

smart_control/notebooks/SAC_Demo.ipynb
s2t2: JK I want to prioritize merging this PR first, before #23

s2t2 (approved):
Looks good to me. Resolves #13

PR #39, opened 2025-05-02
Pre-submit Checks: Undefined Variables

### LINKED ISSUE #38: Fix unknown reference to `reader_utils_google` module
In the "smart_control/utils/environment_utils.py" file, when initializing the `RecordIoReader` class, we see a reference to a module called `reader_lib_google` that has not been imported. 

In the `get_zones_from_config` function:

https://github.com/google/sbsim/blob/44921e52c49ee9a43fae854e124f2686d8b42078/smart_control/utils/environment_utils.py#L67

In the `get_devices_from_config` function:

https://github.com/google/sbsim/blob/44921e52c49ee9a43fae854e124f2686d8b42078/smart_control/utils/environment_utils.py#L78

This code was introduced in [this commit](https://github.com/google/sbsim/commit/3c03b9bb19550ff06c7fa200c659ed43285ed441).

Let's figure out what this reference needs to be changed to.

NOTES:

  + We do have a local "smart_control/utils/reader_lib.py" module in this repo, but it doesn't define the referenced `RecordIoReader` class.

+ In the Google codebase, we do see a similar import that has been aliased, which is likely related:

```py
from google3.corp.ml.smart_buildings.smart_control.utils import reader_lib as reader_lib_google 
```

So maybe we should define the `RecordIoReader` class in this local "smart_control/utils/environment_utils.py" file, and import the class from there.

UPDATE: Or actually since neither of these functions are currently called, we can remove them altogether.

s2t2:
Update files to make pre-submit checks pass, so we can merge the code into Google.

Addresses Undefined Variable -related linting checks: 

  + undefined-variable (E0602)

Also fixes #38.

### DIFF
--- a/smart_control/utils/environment_utils.py
+++ b/smart_control/utils/environment_utils.py
@@ -1,11 +1,8 @@
 """Reinforcement learning environment utility functions."""
 
-from typing import Sequence
-
 import gin
 import pandas as pd
 
-from smart_control.proto import smart_control_building_pb2
 from smart_control.proto import smart_control_normalization_pb2
 from smart_control.utils import bounded_action_normalizer
 
@@ -20,18 +17,6 @@ def to_timestamp(date_str: str) -> pd.Timestamp:
 def local_time(time_str: str) -> pd.Timedelta:
   """Utilty macro for gin config."""
   return pd.Timedelta(time_str)
-
-
-@gin.configurable
-def enumerate_zones(
-    n_building_x: int, n_building_y: int
-) -> Sequence[tuple[int, int]]:
-  """Utilty macro for gin config."""
-  zone_coordinates = []
-  for x in range(n_building_x):
-    for y in range(n_building_y):
-      zone_coordinates.append((x, y))
-  return zone_coordinates
 
 
 @gin.configurable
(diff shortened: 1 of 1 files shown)


smart_control/utils/environment_utils.py
```diff
@@ -56,25 +41,3 @@ def set_action_normalization_constants(
       min_normalized_value,
       max_normalized_value,
   )
-
-
-@gin.configurable
-def get_zones_from_config(
```
s2t2, line 62: This `get_zones_from_config` function is unused anywhere in the codebase. Removing it also fixes an undefined variable reference error, which was the original goal.

smart_control/utils/environment_utils.py
```diff
@@ -56,25 +41,3 @@ def set_action_normalization_constants(
...
-) -> Sequence[smart_control_building_pb2.ZoneInfo]:
-  """Loads up the zones as a gin macro."""
-  with gin.unlock_config():
-    reader = reader_lib_google.RecordIoReader(input_dir=configuration_path)
-    zone_infos = reader.read_zone_infos()
-    return zone_infos
-
-
-@gin.configurable
-def get_devices_from_config(
```
s2t2, line 73: This `get_devices_from_config` function is unused anywhere in the codebase. Removing it also fixes an undefined variable reference error, which was the original goal.

smart_control/utils/environment_utils.py
```diff
@@ -22,18 +19,6 @@ def local_time(time_str: str) -> pd.Timedelta:
   return pd.Timedelta(time_str)
 
 
-@gin.configurable
-def enumerate_zones(
```
s2t2, line 26: This `enumerate_zones` function is also unused anywhere in the codebase, so we are removing it while we're here.

Issue #53, opened 2025-05-13
Spell Checking


s2t2:
We see lots of spelling issues in docstrings and variable names throughout the repo.

It would be ideal to implement some automated spell checking mechanisms. For example using the cSpell extension for VS-Code.

Ideally there are some pre-checks we can do as well.

We may want to also undergo an effort to fix existing spelling issues.

PR #54, opened 2025-05-13
Pre-submit Checks: Google Tests (Ignore Protected Access)

s2t2:
~~Some changes introduced in #46 (regarding changing variable names from protected to public) broke some tests on the Google side.~~ UPDATE: actually this wasn't the cause of the failing tests 😅 

Admittedly, changing all the variable names may have been a bit of a heavy-handed approach. So instead we are reverting all the variables to protected, and ignoring the protected-access `pylint` errors.

Additionally included in this PR are minor revisions which address some spelling issues spotted along the way (#53).

### DIFF
--- a/METADATA
+++ b/METADATA
@@ -13,7 +13,7 @@ third_party {
 
   identifier {
     type: "PyPI"
-    omission_reason: "Does not yet exist on PyPI."
+    omission_reason: "Package doesn't have an upstream repo."
   }
 }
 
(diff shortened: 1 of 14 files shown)


s2t2:
Through #56 we were able to confirm this revision is not necessary. However we still might prefer to ignore the protected Access errors, to minimize diffs in the Google codebase.

s2t2:
Ultimately we are deciding to keep the "protected-access" fixes as-is, instead of ignoring the "protected-access" issues with this PR.

PR #56, opened 2025-05-15
Reorder Commits (v2)

s2t2:
Troubleshooting failing tests. Reordering commits to see how many successive commits we can use until the tests fail.


s2t2:
OK we have successfully identified "Unused References" as the actual culprit. Previously we though the "Protected Access" was the culprit. So we don't need #54 (Ignore Protected Access) after-all.

PR #32, opened 2025-05-01
Pre-submit Checks: Unused References

s2t2:
Update files to make pre-submit checks pass, so we can merge the code into Google.

Addresses Unused Reference-related linting checks: 

  + unused-import (W0611)
  + unused-variable (W0612)
  + unused-argument (W0613)

Notes:

  +  Removes a `_format_action` function from the base `Environment` class that is now-unused by any child classes in this codebase, as a result of removing unused variable references. There is a reference to a `_format_action` function in a child class in the Google codebase, but that class (`HybridActionEnvironment`) defines its own `_format_action` function, so this function in the parent class is confirmed as unnecessary.
  + Removes an unused argument `network_state` from the `call` function in the `_TanhNormalProjectionNetworkWrapper` class. The parent class (`TanhNormalProjectionNetwork` from tf-agents) does not use a `network_state` in its `call` function, so we confirm we can remove this unused arg.

PR #57, opened 2025-05-15
Restore Unused References

s2t2:
Removing unused References in #32 caused Google tests to fail, so let's try to fix the tests by restoring some of the original code that was removed in that PR.

Notes:

  + It was originally thought that because the `HybridActionEnvironment` child class defines its own `_format_action` function that the definition of this function in the parent `Environment` class was unnecessary, because in the parent  class this function was a no-op that [returned the `action` without formatting it](https://github.com/google/sbsim/commit/3e7b4c4da9287da9bc1dae9c5741e1f49c67eec5#diff-cbd32a8e02d1b44ff56eba02dde2e0ca68519fb7581844c61709904a62909359L1224). 
  + However there was also an [invocation](https://github.com/google/sbsim/commit/3e7b4c4da9287da9bc1dae9c5741e1f49c67eec5#diff-cbd32a8e02d1b44ff56eba02dde2e0ca68519fb7581844c61709904a62909359L1259) of the `_format_action` function within the `_step` function of the `Environment` class, and because the `HybridActionEnvironment` class doesn't define its own `_step` function, restoring the invocation in the parent class was necessary to allow the child class to format the actions during each step. 
 + And in order for this invocation in the parent class to be valid we need to also restore the definition of the `_format_action` function within the parent class.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -2,6 +2,10 @@ name: Python CI Build
 
 on:
   push:
+    branches:
+      - master
+      - main
+      - copybara_push
   pull_request:
     branches:
       - master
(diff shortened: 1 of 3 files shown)


s2t2:
Confirm these restorations cause the Google tests to pass again 🎉

Issue #58, opened 2025-05-16
Markdown File Formatting


s2t2:
Right now our formatters (`pyink`, `pylint`) only apply to python files, and they are not recommended for use with markdown files.

We need to check for and fix "line-too-long" errors in markdown files, in order to merge the code into Google codebase.

So we should set up some automated checks for these errors.

It would be nice to integrate markdown formatting checks using a VS Code extension, as well as a pre-commit hook.

PR #60, opened 2025-05-16
Pre-submit Checks: Ignore Markdown Formatting

s2t2:
Ignores markdown formatting checks for "line-too-long" errors that appear in markdown files. Addresses https://github.com/google/sbsim/issues/58


s2t2:
For some reason the presence of comments to ignore certain long lines is actually disabling all Lint checks altogether. 🤷

PR #61, opened 2025-05-16
Pre-submit Checks: METADATA

s2t2:
Fixes line-too-long error in METADATA file.


s2t2:
Changes to the METADATA file get overridden by the Copybara sync and are not respected. These changes would need to happen on the Google side.

PR #62, opened 2025-05-16
Pre-submit Checks: Exempt Markdown Files

### LINKED ISSUE #58: Markdown File Formatting

s2t2:
Closes #58

### DIFF
--- a/METADATA
+++ b/METADATA
@@ -45,7 +45,8 @@ presubmit: {
 
 presubmit: {
   path_expression: "//depot/google3/third_party/py/smart_buildings/..."
-  path_expression_exclusion: "....md"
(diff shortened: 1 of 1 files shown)


s2t2:
Changes to the METADATA file get overridden by the Copybara sync and are not respected. These changes would need to happen on the Google side.

PR #77, opened 2025-05-24
Add Dockerfile

ozanbarism:
- Added a Dockerfile for setting up the environment on OS that are not Linux, since some libraries require linux. 
- Added a short description of how to set up the docker container.


Dockerfile
```diff
@@ -0,0 +1,83 @@
+# Use an x86_64 specific base image
+FROM --platform=linux/amd64 ubuntu:20.04
+
+# Set up environment
+ENV DEBIAN_FRONTEND=noninteractive
+ENV PYTHONUNBUFFERED=1
+WORKDIR /workspace
+
+# Install system dependencies
+RUN apt-get update && apt-get install -y \
```
s2t2, line 10: Let's consider splitting up this RUN command to allow Docker to cache more effectively. Comment from Gemini below...

Issue: The `rm -rf /var/lib/apt/lists/*` is at the end of a very long RUN command. If any of the previous `apt-get install` commands fail, this cleanup step won't be reached, leaving potentially large caches.

Suggestion: Split the `apt-get update` and `apt-get install` into separate RUN commands, but keep the `rm -rf` with the last `apt-get install` that modifies packages. This helps with caching: if `apt-get` update changes, only that layer and subsequent ones are rebuilt.

```
RUN apt-get update && \
    apt-get install -y software-properties-common && \
    add-apt-repository ppa:deadsnakes/ppa && \
    apt-get update

RUN apt-get install -y \
    python3.10 python3.10-venv python3.10-distutils \
    protobuf-compiler ffmpeg git curl build-essential libffi-dev libssl-dev \
    pkg-config libhdf5-dev openjdk-11-jdk wget gnupg unzip && \
    rm -rf /var/lib/apt/lists/*
```

Dockerfile
```diff
@@ -0,0 +1,83 @@
...
+    software-properties-common && \
+    add-apt-repository ppa:deadsnakes/ppa && \
+    apt-get update && apt-get install -y \
```
s2t2, line 13: Let's consider removing the duplicate `apt-get update` here, perhaps as part of an effort to split the RUN command into two smaller ones (see above).

Dockerfile
```diff
@@ -0,0 +1,83 @@
...
+    python3.10 python3.10-venv python3.10-distutils \
```
s2t2, line 14: Is it possible for us to use Python 3.11 here, just to get a bit ahead?

Dockerfile
```diff
@@ -0,0 +1,83 @@
...
+# Clone the repository to a temporary location
+RUN git clone https://github.com/google/sbsim.git /tmp/sbsim
+
+# Configure poetry
+RUN cd /tmp/sbsim && \
+    poetry config virtualenvs.create false && \
+    poetry config installer.max-workers 10
+
+# Install base dependencies first
+RUN /opt/venv/bin/pip install \
```
s2t2, line 51: Here we are using Pip to install these dependencies, and then later using Poetry to install additional dependencies. However it would be nice to consolidate the installation approach to rely on Poetry as much as possible. 

For example, it looks like Poetry is already installing `dm-reverb`, so we could perhaps remove that from the Pip installation step. 

Are the rest of these required for running notebooks? If so, let's consider adding `jupyter` and any other notebook-related dependencies to the dev dependencies group in the "pyproject.toml" file.

s2t2 (commented):
@ozanbarism Thanks for adding this Dockerfile! 🐳 This is an awesome addition that will significantly help expand adoption to include Mac OS users.

I'm not the most experienced with Docker, so I asked Gemini to review, and relayed some of the comments it provided. I also added some comments of my own. Let me know your thoughts about these comments.

Gemini seemed to advocate for splitting up the large RUN command for all the Linux installations, but ChatGPT seemed to prefer to keep the single large RUN command 🤷 

If the current Dockerfile is working for you, we can definitely merge this PR as-is. Alternatively, if we want to make some changes, let me know if you would prefer for us to make the changes in this PR, or make an issue to "Optimize the Dockerfile" as a follow-on PR.

s2t2:
Oh, also, I ran the Dockerfile contents through https://hadolint.github.io/hadolint/ and it provides the following recommendations as well for us to consider:

  + Pin versions in apt get install. Instead of `apt-get install <package>` use `apt-get install <package>=<version>` -- [more info](https://github.com/hadolint/hadolint/wiki/DL3008)
  + Use WORKDIR to switch to a directory, instead of cd -- [more info](https://github.com/hadolint/hadolint/wiki/DL3003)
  + Set the SHELL option -o pipefail before RUN with a pipe in it -- [more info](https://github.com/hadolint/hadolint/wiki/DL4006)
  + This ln has no destination. Check the arguments, or specify '.' explicitly. -- [more info](https://github.com/koalaman/shellcheck/wiki/SC2226)
  + Either use Wget or Curl but not both -- [more info](https://github.com/hadolint/hadolint/wiki/DL4001)
  + echo may not expand escape sequences. Use printf. -- [more info](https://github.com/koalaman/shellcheck/wiki/SC2028)

s2t2 (approved):
Looks good to me. I will add an issue called "Optimize Docker file" if anyone (especially someone more experienced with Docker) would like to follow up with some Dockerfile improvements and optimizations later.

UPDATE: see https://github.com/google/sbsim/issues/80

PR #87, opened 2025-06-03
Dataset Tests

s2t2:
Tests the `SmartBuildingsDataset` class defined in "smart_control/dataset/dataset.py".

Details:

  + Adds tests for the dataset download process, as well as to describe the structure of the resulting data. This includes high fidelity end-to-end tests for dataset downloading. Skips these tests by default, because it takes around two minutes to download the data and we don't want to slow down the build. However we can run these tests periodically as desired by passing a `TEST_DATASET_DOWNLOAD` environment variable. The presence of these tests, even if not always run, is helpful for reading and understanding the behavior of the code.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,7 +14,9 @@ __pycache__/
 # data downloads:
 sb1.zip
 sb1/
+
 data/*
+!data/.gitkeep
 data/sb1.zip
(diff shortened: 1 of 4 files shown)


s2t2:
Superceded by #97

PR #93, opened 2025-06-06
Mac OS Setup Instructions

s2t2:
Adds instructions for setting up the repository on Mac OS.

Details:

  + Uses homebrew to install system libraries on Mac, and Anaconda to manage the virtual environment.

  + Pins specific versions of system library dependencies (`protoc` version 3.21.12 and `ffmpeg` version 7.1.1), to prevent differences between setups. Also pins `poetry` package to version 3.1.2.

  + Updates the "Makefile" to use poetry-based commands instead of virtualenv-specific commands that won't work if a Mac user is using Anaconda instead.

  + Splits the setup instructions into three separate system-specific documents (Mac vs Linux vs Docker).

  + Updates the GitHub Actions workflow file to run tests against Mac OS as well.

  + Fixes a test that was passing on Linux and newer versions of Mac (ARM64), but failing on intel-based Macs (x86_64) due to slight differences in floating point math. 

  + We are noticing an issue when trying to use the `dm-reverb` package on Mac. We attempted to overcome the issue by conditionally installing `dm-reverb-macos` on Mac, however that doesn't work either. We are deciding to move forward with merging this Mac OS setup, despite the `dm-reverb` issues, because it only affects a small part of the codebase (the RL replay buffers). We are adding a note to the Mac OS setup instructions about this, and creating a separate issue to follow up with fixing the `dm-reverb` installation on Mac. #102


s2t2:
Interesting. Seeing a test fail on Mac that passes on Linux ("smart_control/utils/regression_building_utils_test.py"):

```sh
>     self.assertDictEqual(expected_feature_map, feature_map)
E     AssertionError: {'timestamp': Timestamp('2021-01-12 00:10:00+0000', tz='UTC'), ('d0', 's0'): 294.5, ('dow', 'cos_000'): 0.6234898018587336, ('dow', 'sin_000'): 0.7818314824680298, ('hod', 'cos_000'): 0.9990482215818578, ('hod', 'sin_000'): 0.043619387365336} != {'timestamp': Timestamp('2021-01-12 00:10:00+0000', tz='UTC'), ('d0', 's0'): 294.5, ('dow', 'cos_000'): 0.6234898018587336, ('dow', 'sin_000'): 0.7818314824680297, ('hod', 'cos_000'): 0.9990482215818578, ('hod', 'sin_000'): 0.043619387365336}
E     repr() of differing entries:
E     ('dow', 'sin_000'): 0.7818314824680298 != 0.7818314824680297

smart_control/utils/regression_building_utils_test.py:150: AssertionError
```

The values are very very similar. I wonder if calculations are slightly different on Linux vs Mac. We might consider using almost equal assertions.

s2t2:
TODO: pin the poetry version to 2.1.2

s2t2:
Looks like we are running into issues with trying to use`dm-reverb` on Mac OS.

https://github.com/google-deepmind/reverb/issues/11

https://github.com/google-deepmind/reverb/pull/128

s2t2:
We are deciding to move forward with merging this Mac OS setup, despite the `dm-reverb` issues, which only affect a small part of the codebase (the RL replay buffers). We will add a note to the Mac OS setup instructions, and create a separate issue to follow up with fixing the `dm-reverb` installation on Mac.

PR #94, opened 2025-06-06
LBNL - Interior Radiative Heat Transfer

ecosang:
Enables the Simulator Building to account for interior radiative heat transfer.

Details:

  + Adds radiative heat transfer utility functions in "smart_control/simulator/building_radiation_utils.py". Adds documentation for these functions, including links to source references and details about the equations implemented. Adds tests for these utility functions. 

  + Implements longwave interior radiative heat transfer in the `FloorPlanBasedBuilding`. Passes various radiation properties, along with a new `include_radiative_heat_transfer` boolean flag to determine whether the radiative heat transfer calculations should be enabled or not (disabled by default).  Implements view factor calculations for modeling long wave radiation, currently using the "scriptF" algorithm, but provides future extension to use the "CarrollMRT" algorithm. See: [Long Wave Radiation Exchange Among Zone Surfaces](https://bigladdersoftware.com/epx/docs/9-6/engineering-reference/inside-heat-balance.html#lw-radiation-exchange-among-zone-surfaces). Implements iterative as well as tensorized calculations.

  + Specifies default radiative heat property values for indoor air, indoor walls, and building exteriors.

  + Updates the `Simulator`, when calculating the interior control volume temperature, to conditionally consider radiative heat transfer if the building's `include_radiative_heat_transfer` mode is enabled.

  + Updates the documentation site to show the newly implemented methods. Installs mathjax in the documentation site to enable display of equations in the docstrings.


smart_control/simulator/building_utils_radiation.py
```diff
@@ -0,0 +1,498 @@
...
+) -> np.ndarray:
+  """
+  Calculate view factors between interior walls in the floor plan.
+
+
+  Args:
+      indexed_floor_plan (np.ndarray): 2D array representing the floor plan with indexed values.
+      interior_wall_value (int, optional): Value representing interior walls. Defaults to -3 (constants.INTERIOR_WALL_VALUE_IN_FUNCTION).
+      marked_value (int, optional): Value to mark connected walls. Defaults to -33.
+      air_value (int, optional): Value representing air spaces. Defaults to 0.
```
s2t2: Here `air_value` is in the docstring for the `get_VF` function, but it does not appear in the function signature and is not currently used in the function. So I will remove it. However if this is not desired, please let me know.

s2t2:
@ecosang I have renamed the radiation utils file to "smart_control/simulator/building_radiation_utils.py" and updated the documentation in this file.

I configured mathjax so we can now include equation formatting in the docstrings (see example in `calculate_IFAinv` function).

If you install the docs dependencies with `poetry install --with docs` and then run `make docs`, you should be able to view the documentation for this file at http://localhost:8000/sbsim/api/simulator/utils/#smart_control.simulator.building_radiation_utils. Feel free to update the docs as desired.

s2t2:
@ecosang do you think there are some basic unit tests we could add for the "smart_control/simulator/building_radiation_utils.py" file? 

It would be nice to have some tests that exemplify the functionality of the radiation utility functions, using very simple inputs, to let us know what the expected outputs look like.

We could add the tests to a new file called "smart_control/simulator/building_radiation_utils_test.py".

smart_control/simulator/building.py
```diff
@@ -28,6 +29,16 @@ class MaterialProperties:
   density: float
 
 
+@gin.configurable
+@dataclasses.dataclass
+class RadiationProperties:
```
s2t2, line 34: We might consider moving this class to the radiation utils file, because it is radiation-related.

s2t2: However the material properties data class is currently in this file, so there is justification for us keeping it here for now.

s2t2:
> @ecosang do you think there are some basic unit tests we could add for the "smart_control/simulator/building_radiation_utils.py" file?
> 
> It would be nice to have some tests that exemplify the functionality of the radiation utility functions, using very simple inputs, to let us know what the expected outputs look like.
> 
> We could add the tests to a new file called "smart_control/simulator/building_radiation_utils_test.py".

I added an example test in "smart_control/simulator/building_radiation_utils_test.py". Feel free to update / revise. Let me know what you think.

smart_control/simulator/building.py
```diff
@@ -757,6 +782,78 @@ def __init__(
...
+          epsilon=0.0, alpha=0.0, tau=0.0
+      )
+    if inside_air_radiative_properties is None:
+      inside_air_radiative_properties = RadiationProperties(
+          epsilon=0.0, alpha=0.0, tau=0.0
+      )
+
+    # emissivity
+    self._epsilon = _assign_interior_and_exterior_values(
+        exterior_walls=exterior_walls,
```
s2t2: for epsilon, alpha, and tau: do we want to use the `exterior_walls` and `interior_walls` (as-is), or the enlarged interior walls and enlarged exterior walls available at `self._interior_walls` and `self._exterior_walls`?

ecosang: @s2t2 Thanks for the catch. Do you know what's the difference between `enlarge_exterior_walls` stuff? Initially, 
The plan is like `[outdoor_air, exteiror_wall, interior_wall, air ,...]`  and then the the `enlarge` removes `interior_wall` when the `interior_wall` neighbors with `exterior_walls`. Do you know why it is used? .
But for the radiative heat transfer, we need to know all the interior wall information, so I use non-enlarged version.

ecosang: For example, I have -1 for outdoor air, -2 exterior wall, and -3 for interior walls.
```
self.indexed_floor_plan
array([[-1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1],
       [-1, -2, -2, -2, -2, -2, -2, -2, -2, -2, -2, -1],
       [-1, -2, -3, -3, -3, -3, -3, -3, -3, -3, -2, -1],
       [-1, -2, -3,  0,  0,  0,  0,  0,  0, -3, -2, -1],
       [-1, -2, -3,  0,  0,  0,  0,  0,  0, -3, -2, -1],
       [-1, -2, -3,  0,  0,  0,  0,  0,  0, -3, -2, -1],
       [-1, -2, -3,  0,  0,  0,  0,  0,  0, -3, -2, -1],
       [-1, -2, -3,  0,  0,  0,  0,  0,  0, -3, -2, -1],
       [-1, -2, -3,  0,  0,  0,  0,  0,  0, -3, -2, -1],
       [-1, -2, -3,  0,  0,  0,  0,  0,  0, -3, -2, -1],
       [-1, -2, -3,  0,  0,  0,  0,  0,  0, -3, -2, -1],
       [-1, -2, -3, -3, -3, -3, -3, -3, -3, -3, -2, -1],
       [-1, -2, -3,  0,  0,  0,  0,  0,  0, -3, -2, -1],
       [-1, -2, -3,  0,  0,  0,  0,  0,  0, -3, -2, -1],
       [-1, -2, -3,  0,  0,  0,  0,  0,  0, -3, -2, -1],
       [-1, -2, -3,  0,  0,  0,  0,  0,  0, -3, -2, -1],
       [-1, -2, -3,  0,  0,  0,  0,  0,  0, -3, -2, -1],
       [-1, -2, -3,  0,  0,  0,  0,  0,  0, -3, -2, -1],
       [-1, -2, -3,  0,  0,  0,  0,  0,  0, -3, -2, -1],
       [-1, -2, -3,  0,  0,  0,  0,  0,  0, -3, -2, -1],
       [-1, -2, -3, -3, -3, -3, -3, -3, -3, -3, -2, -1],
       [-1, -2, -2, -2, -2, -2, -2, -2, -2, -2, -2, -1],
       [-1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1]])
```
The enlarged version rconvert -3 to -2 when -3 is neighboring to the -2

```
self._exterior_walls
array([[ 0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0],
       [ 0, -2, -2, -2, -2, -2, -2, -2, -2, -2, -2,  0],
       [ 0, -2, -2, -2, -2, -2, -2, -2, -2, -2, -2,  0],
       [ 0, -2, -2,  0,  0,  0,  0,  0,  0, -2, -2,  0],
       [ 0, -2, -2,  0,  0,  0,  0,  0,  0, -2, -2,  0],
       [ 0, -2, -2,  0,  0,  0,  0,  0,  0, -2, -2,  0],
       [ 0, -2, -2,  0,  0,  0,  0,  0,  0, -2, -2,  0],
       [ 0, -2, -2,  0,  0,  0,  0,  0,  0, -2, -2,  0],
       [ 0, -2, -2,  0,  0,  0,  0,  0,  0, -2, -2,  0],
       [ 0, -2, -2,  0,  0,  0,  0,  0,  0, -2, -2,  0],
       [ 0, -2, -2,  0,  0,  0,  0,  0,  0, -2, -2,  0],
       [ 0, -2, -2, -2,  0,  0,  0,  0, -2, -2, -2,  0],
       [ 0, -2, -2,  0,  0,  0,  0,  0,  0, -2, -2,  0],
       [ 0, -2, -2,  0,  0,  0,  0,  0,  0, -2, -2,  0],
       [ 0, -2, -2,  0,  0,  0,  0,  0,  0, -2, -2,  0],
       [ 0, -2, -2,  0,  0,  0,  0,  0,  0, -2, -2,  0],
       [ 0, -2, -2,  0,  0,  0,  0,  0,  0, -2, -2,  0],
       [ 0, -2, -2,  0,  0,  0,  0,  0,  0, -2, -2,  0],
       [ 0, -2, -2,  0,  0,  0,  0,  0,  0, -2, -2,  0],
       [ 0, -2, -2,  0,  0,  0,  0,  0,  0, -2, -2,  0],
       [ 0, -2, -2, -2, -2, -2, -2, -2, -2, -2, -2,  0],
       [ 0, -2, -2, -2, -2, -2, -2, -2, -2, -2, -2,  0],
       [ 0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0]], dtype=int16)
```
while I just use the exterior wall information
```
exterior_walls
array([[ 0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0],
       [ 0, -2, -2, -2, -2, -2, -2, -2, -2, -2, -2,  0],
       [ 0, -2,  0,  0,  0,  0,  0,  0,  0,  0, -2,  0],
       [ 0, -2,  0,  0,  0,  0,  0,  0,  0,  0, -2,  0],
       [ 0, -2,  0,  0,  0,  0,  0,  0,  0,  0, -2,  0],
       [ 0, -2,  0,  0,  0,  0,  0,  0,  0,  0, -2,  0],
       [ 0, -2,  0,  0,  0,  0,  0,  0,  0,  0, -2,  0],
       [ 0, -2,  0,  0,  0,  0,  0,  0,  0,  0, -2,  0],
       [ 0, -2,  0,  0,  0,  0,  0,  0,  0,  0, -2,  0],
       [ 0, -2,  0,  0,  0,  0,  0,  0,  0,  0, -2,  0],
       [ 0, -2,  0,  0,  0,  0,  0,  0,  0,  0, -2,  0],
       [ 0, -2,  0,  0,  0,  0,  0,  0,  0,  0, -2,  0],
       [ 0, -2,  0,  0,  0,  0,  0,  0,  0,  0, -2,  0],
       [ 0, -2,  0,  0,  0,  0,  0,  0,  0,  0, -2,  0],
       [ 0, -2,  0,  0,  0,  0,  0,  0,  0,  0, -2,  0],
       [ 0, -2,  0,  0,  0,  0,  0,  0,  0,  0, -2,  0],
       [ 0, -2,  0,  0,  0,  0,  0,  0,  0,  0, -2,  0],
       [ 0, -2,  0,  0,  0,  0,  0,  0,  0,  0, -2,  0],
       [ 0, -2,  0,  0,  0,  0,  0,  0,  0,  0, -2,  0],
       [ 0, -2,  0,  0,  0,  0,  0,  0,  0,  0, -2,  0],
       [ 0, -2,  0,  0,  0,  0,  0,  0,  0,  0, -2,  0],
       [ 0, -2, -2, -2, -2, -2, -2, -2, -2, -2, -2,  0],
       [ 0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0]])
```

ecosang:
@s2t2 I've added several updates.

Features
- Add include_radiative heat transfer flag in building  (default False) (and subsequently, the calculation skips not enable).
- View_factor_method flag in building (default False)
- Add scriptF view factor calculation.

Test cases
- test cases for view-factor related geometry in building_test.py
- test cases for building_radiation_utils.py
- test case for simulator with radiative heat transfer. 
 
Also, revert other simulator_flexible_floor_plan_test cases to its original version as they don't include radiative heat transfer at this moment. Intesad, I added one `test_update_temperature_estimates_return_value_with_radiative_heat_transfer`

smart_control/simulator/building.py
s2t2: I would say use whichever makes most sense to you. I was going through the code and saw the different versions (original vs enlarged) and just wanted to bring your attention.

s2t2:
@ecosang this looks like some awesome progress, thank you! I will avoid making any new commits this week to let you finish making any final touches on this new iteration.  Let's discuss this next round of updates during our meeting on Friday.

smart_control/simulator/building_test.py
```diff
@@ -1714,6 +1716,8085 @@ def test_stochastic_convection_simulator_shuffle_max_dist(
...
+        [-1, -2, -3, 0, 0, 0, 0, 0, 0, -3, -2, -1],
+        [-1, -2, -3, 0, 0, 0, 0, 0, 0, -3, -2, -1],
+        [-1, -2, -3, 0, 0, 0, 0, 0, 0, -3, -2, -1],
+        [-1, -2, -3, -3, -3, -3, -3, -3, -3, -3, -2, -1],
+        [-1, -2, -2, -2, -2, -2, -2, -2, -2, -2, -2, -1],
+        [-1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1],
+    ])
+
+    expected_interior_wall_vf = np.array([
+        [
```
s2t2: don't you love it when the formatter puts a single number per line, for hundreds / thousands of lines 😅 

we might prefer to skip formatting here (`fmt: off`) to improve readability. like:

```py
# pylint:disable=line-too-long
# fmt: off

all the numbers [
    [1,1,1,1,1,1,1,1,1,1,1 ........... ],
    [2,2,2,2,2,2,2,2,2,2,2,2,...... ],
   etc,
]
# pylint:enable=line-too-long
# fmt: on
```

s2t2 (commented):
This PR is looking good generally. Appreciate the additional tests, and the conditionals to turn on Radiative Heat Transfer mode.

smart_control/simulator/building.py
```diff
@@ -665,14 +714,25 @@ def __init__(
...
+      inside_air_radiative_properties: RadiationProperties for interior air.
+      inside_wall_radiative_properties: RadiationProperties for interior walls.
+      building_exterior_radiative_properties: RadiationProperties for building's
+        exterior.
+      include_radiative_heat_transfer: bool to note whether to include radiative
+        heat transfer.
+      view_factor_method: str to note the method to use for view factors.
+        Either "ScriptF" or "CarrollMRT". See
+        https://bigladdersoftware.com/epx/docs/9-6/engineering-reference/
+        inside-heat-balance.html#lw-radiation-exchange-among-zone-surfaces
```
s2t2: we can use a markdown link here to keep the link intact, and it should not violate line too long.

ecosang:
@s2t2 updated docs with equations. Also, making it work with no-interior wall case.

filename.csv
s2t2, line 1: @ecosang I see a CSV file here, but didn't see a reference to it. Should we remove it?

ecosang: ah sorry, that one should be removed. my mistake.

s2t2:
> @s2t2 updated docs with equations. Also, making it work with no-interior wall case.

Nice! Those equations are super helpful 🙌 And glad to hear you were able to chase down the case with no interior walls.

I just updated formatting of those super long lists to be line per list instead of line per number. I also reorganized the building tests in a minor way (putting expected arrays inside their respective sub-tests).

filename.csv
s2t2: OK I will remove it now. All good!

smart_control/simulator/building.py
```diff
@@ -28,6 +29,22 @@ class MaterialProperties:
   density: float
 
 
+@gin.configurable
+@dataclasses.dataclass
+class RadiationProperties:
```
s2t2, line 34: @ecosang For the `RadiationProperties`, I was wondering, what would be a valid or usual range of values? Are we exclusively in the [0,1] space? I was hoping to add some more context to the documentation about what the absorptivity, emissivity, and transmittance values could be, and what would be some common values in normal situations. Thanks!

ecosang: oh thanks. they are all in [0,1].

smart_control/simulator/building.py
```diff
@@ -623,6 +655,34 @@ class FloorPlanBasedBuilding(BaseSimulatorBuilding):
...
+    IFAinv: matrix representing the inverse of the IFA matrix of the nodes of
+      the building. Used only for calculating radiative heat transfer.
+
+      The longwave radiation ($q_{lwx}$) is calculated as:
+
+      $$q_{lwx} = \\sigma(I-F)\\tilde{A}_{inv}T^4$$
+
+      Where the term $(I-F)\\tilde{A}_{inv}$ can be pre-calculated as:
+
+      $$IFA_{inv} = (I-F)\\tilde{A}_{inv}$$
```
s2t2, line 789: FYI - equation as displayed in the documentation site:

Image

smart_control/simulator/building_radiation_utils.py
```diff
@@ -0,0 +1,841 @@
...
+    for j in range(n):
+      A[i, j] = (I[i, j] - (1 - epsilon[i]) * F[i, j]) / epsilon[i]
+  return np.linalg.inv(A)
+
+
+def calculate_IFAinv(F: np.ndarray, A_inv: np.ndarray) -> np.ndarray:
+  r"""
+  Calculates the $IFA_{inv}$ matrix.
+
+  $$IFA_{inv} = (I - F) \tilde{A}^{-1}$$
```
s2t2, line 53: FYI - equation as displayed in the documentation site:

Image

smart_control/simulator/building_radiation_utils.py
```diff
@@ -0,0 +1,841 @@
...
+
+
+def net_radiative_heatflux_function_of_T(
+    T: np.ndarray, IFAinv: np.ndarray
+) -> np.array:
+  r"""
+  Calculates the net radiative heat flux and radiosity for all surfaces given
+    surface temperatures.
+
+  Equations:
```
s2t2, line 79: FYI - equations as displayed in the documentation site:

Image

Image

Image

smart_control/simulator/building.py
```diff
@@ -757,8 +837,116 @@ def __init__(
...
+          self.interior_wall_mask,
+          constants.INTERIOR_WALL_VALUE_IN_FUNCTION,
+          view_factor_method=view_factor_method,
+      )
+
+      # radiative properties
+      # by default, all radiative properties are 0.0
+      inside_wall_radiative_properties = (
+          inside_wall_radiative_properties
+          or RadiationProperties(epsilon=0.0, alpha=0.0, tau=0.0)
```
s2t2: Instead of defaulting these properties to zero, we would like to be able to specify some "reasonable" defaults for epsilon, alpha, and tau for the inside wall properties, exterior properties, and inside air properties.

It would be nice to be able to say that, for example, "a glass wall would be equivalent to values X, Y, and Z".  

Is there a place we can look these up? It would be nice to link to more information, and perhaps include a few different options in a lookup table in the code and/or docs.

smart_control/simulator/building_radiation_utils_test.py
```diff
@@ -0,0 +1,296 @@
...
+        [ -1,  -2, -2,  -2,  -2,  -2,  -2,  -2,  -2,  -2,  -2,  -1],
+        [ -1,  -1, -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1]]
+      )
+
+
+
+    expected_result_23 = \
+      np.array([[ -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
+       [ -1,  -2,  -2,  -2,  -2,  -2,  -2,  -2,  -2,  -2,  -2,  -1],
+       [ -1,  -2,  -3, -67, -34, -34, -34, -34, -34,  -3,  -2,  -1],
```
s2t2, line 230: We have a question about the meaning of these values: for example, what does -67 represent and what does -34 represent? 

It would be nice to have some documentation about this. Apologies if we overlooked this somewhere. Perhaps we can set some constant values in the code.

s2t2 (changes requested):
@ecosang, thanks again, this PR is looking awesome! 

@vorhersager and I just did a final review and we have two final comments / questions (see below). 

We would appreciate your feedback on these final comments, then we can merge this PR.

After we merge this PR we can implement a follow-up PR with evaluation methods to see the effects of the internal radiative heat transfer on the zone air temperatures over a period of a few hours.

smart_control/simulator/building_radiation_utils_test.py
ecosang: i am out of town now. back in 2weeks later. will update near that time. Sorry!

s2t2:
@ecosang some follow ups as we get closer to merging this PR:

1. Thank you for adding more information about the **floorplan values**! One note is that I recently spoke with a previous contributor and he said we are reserving negative numbers for outside / exterior. Sorry this hadn't been documented anywhere (I wasn't aware of it myself). To stick with the convention, we might want to make those negative numbers positive instead. Let me know if you are able to implement the change, or if you would prefer I do it. 

2. Thanks for adding the **tensorized approach**! 🙌 Do you think this is complete at this time? If complete we can include in this PR, otherwise if there is more work we might consider a separate PR. Let me know your preference.

3. For the default radiative property values for various known materials... I asked Gemini deep research for some more info and it provided a response which I summarized below. See also the Google doc I shared with you, which provides a copy of the entire response. What do you think about these proposed values? Are they verifiable / defensible? If these values look reasonable to you and you are confident in them, let me know and I would be happy to make some updates to the code.


---

## Inside Air

Based on the characterization of air as a non-participating medium for the purpose of a simplified building simulator, the recommended default values for the `RadiationProperties` for `inside_air_radiative_properties` are as follows:

  + alpha (absorptivity): 0.0
  + epsilon (emissivity): 0.0
  + tau (transmittance): 1.0
  
These values represent a material that is fully transparent to radiation, meaning it neither absorbs nor emits energy and allows all radiation to pass through unimpeded.

## Inside Wall

The following table provides a summary of recommended default values for common interior building materials. For these opaque materials, it can be assumed that transmittance is negligible and that absorptivity is equal to emissivity due to Kirchhoff's Law.

Material Type | Recommended ϵ Range (Long-wave) | Recommended α Range (Long-wave) | Recommended τ Value (Long-wave) | Source Reference
-- | -- | -- | -- | --
Painted Drywall & Plaster | 0.90-0.98 | 0.90-0.98 | 0.0 | 18
Concrete & Masonry (Rough) | 0.90-0.96 | 0.90-0.96 | 0.0 | 18
Wood (Sanded, Planed) | 0.85-0.95 | 0.85-0.95 | 0.0 | 18
Common Building Surfaces (General) | 0.90-0.98 | 0.90-0.98 | 0.0 | 23

Recommended defaults for `inside_wall_radiative_properties`: 

  + alpha: 0.9
  + epsilon: 0.9
  + tau: 0.0

Note: This is a robust default because most interior surfaces are covered in materials with very high emissivities, like paint or plaster. The precise color of the paint matters more for solar radiation (visible light) than for long-wave thermal radiation.


## Building Exterior

This table outlines the distinct radiative properties for exterior materials across both the short-wave (solar) and long-wave (thermal) spectrums. For opaque materials, transmittance (τ) is zero.

Material Type & Color | Solar Absorptivity (αsolar​) Range | Long-wave Emissivity (ϵlong−wave​) Range | Source Reference
-- | -- | -- | --
Paints |   |   |  
Light-colored (White, Cream) | 0.10-0.30 | 0.90-0.96 | 18
Dark-colored (Black, Brown) | 0.85-0.95 | 0.90-0.96 | 18
Aluminum Paint | 0.27-0.67 | 0.27-0.67 | 18
Masonry |   |   |  
Rough Concrete | 0.60-0.80 | 0.90-0.94 | 1
Rough Red Brick | 0.65-0.85 | 0.90-0.95 | 1
Rough Limestone/Stone | 0.60-0.80 | 0.90-0.96 | 9
Siding |   |   |  
Wood Siding (Painted) | Varies with color | 0.90-0.95 | 19
Oxidized Metal Siding | 0.70-0.90 | 0.70-0.90 | 1
Polished Metal Siding | 0.05-0.20 | 0.05-0.20 | 1

For a general-purpose building exterior model, especially for opaque walls, a high emissivity value is a safe bet. Recommended values for `building_exterior_radiative_properties`: 

  + alpha: 0.9
  + epsilon: 0.9
  + tau: 0.0

Note: This assumes an opaque exterior surface like brick, concrete, or painted siding. If the model includes windows, they should be handled as a separate material with their own unique properties as described above.

### Glass Walls

Glass is a unique building material because it is not opaque and its radiative properties are highly spectral-dependent.[15] Standard clear glass exhibits a "greenhouse effect," allowing most short-wave solar radiation to pass through unimpeded, while being largely opaque to and an efficient emitter of long-wave thermal radiation.[15] This dual behavior makes accurate modeling of glass surfaces critical for simulating heat transfer through the building envelope.

The following table provides a breakdown of recommended radiative property ranges for different types of glass, clearly distinguishing between the two key wavelength spectra.

Glazing Type | Wavelength Spectrum | α Range | ϵ Range | τ Range | Source Reference
-- | -- | -- | -- | -- | --
Standard Clear Glass | Short-wave (Solar) | 0.10-0.20 | 0.10-0.20 | 0.80-0.90 | 11
  | Long-wave (Thermal) | 0.80-0.94 | 0.80-0.94 | 0.0 | 9
Low-E Hard-Coat | Short-wave (Solar) | 0.10-0.30 | 0.10-0.30 | 0.40-0.70 | 29
  | Long-wave (Thermal) | 0.20-0.40 | 0.20-0.40 | 0.0 | 27
Low-E Soft-Coat | Short-wave (Solar) | 0.10-0.25 | 0.10-0.25 | 0.50-0.80 | 30
  | Long-wave (Thermal) | 0.05-0.20 | 0.05-0.20 | 0.0 | 33


## Summary

Normal Gemini:

Material	| Absorptivity (α)	| Emissivity (ϵ)	| Transmittance (τ)
--- | --- | --- | ---
Inside Air	| ~0.0	| ~0.0	| ~1.0
Inside Walls (Plaster, Paint)	| 0.85 - 0.95	| 0.85 - 0.95	| 0.0
Building Exterior (Opaque)	| 0.85 - 0.95	| 0.85 - 0.95	| 0.0
Glass Wall (for Long-Wave IR)	| 0.85 - 0.95	| 0.85 - 0.95	| ~0.0

Deep Research Gemini:

Material Category | Recommended α Range | Recommended ϵ Range | Recommended τ Range | Notes & Assumptions
-- | -- | -- | -- | --
inside_air_radiative_properties | 0.0 | 0.0 | 1.0 | Air is treated as a non-participating medium for radiative transfer.
inside_wall_radiative_properties | 0.90-0.95 | 0.90-0.95 | 0.0 | Assumes a painted or common non-metallic opaque surface; based on long-wave properties.
building_exterior_radiative_properties | 0.90-0.95 | 0.90-0.95 | 0.0 | Assumes a common opaque surface; based on long-wave properties for radiative cooling calculations. Solar properties for this material must be handled separately.
glass_wall_radiative_properties | 0.80-0.94 | 0.80-0.94 | 0.0 | Standard clear glass; based on long-wave properties for internal heat transfer and heat loss.

## References

1 url:
https://www.numberanalytics.com/blog/radiation-in-building-science-essentials

9 url:
https://en.wikipedia.org/wiki/Emissivity

11 url:
https://www.pilkington.com/en/global/knowledge-base/types-of-glass/solar-control-glass/how-it-works

15 url:
https://fenix.ciencias.ulisboa.pt/downloadFile/1126037345801434/SEE-Radiation.pdf

18 url:
https://help.covetool.com/en/articles/5560310-wall-emissivity

19 url:
https://ennologic.com/wp-content/uploads/2018/07/Ultimate-Emissivity-Table.pdf

27 url:
https://www.mornglass.com/thermal-insulation-function-of-low-e-coating.html

29 url:
https://www.nachi.org/low-e-windows.htm

30 url:
https://buildings.lbl.gov/publications/spectrally-selective-glazings

33 url:
https://claytonglass.co.uk/news/the-advantages-and-disadvantages-of-low-e-glass/

ecosang:
@s2t2
Hi Michael,

Thanks for the reply. First of all, I have a deadline next Wednesday, so I may work on this sometime next week.

**1. Notation**
Yeah, that's what I previously emailed Judah about, and I asked several times regarding the notations as I don't want to use my own. The interior wall is '3', so I used something like '33' to avoid the other reserved numbers. Anyway, when I talked with John yesterday, he said I need to define glazing on my own. '33' and '34' are temporary values. At the same time, I think it is a good idea to have a table to define the codes for the floor plan. When I work on changing it to positive values, I will consider glazings (windows) and other notations. If you have any materials regarding the notation (or coding) for materials, walls, or the floor plan, please share them with me.

**2. Tensorized Calculation**
Yes (😃 I was so worried about it, but it is better than I expected!). I am thinking of adding an example case to double-check it against the iterative approach.
I have added the initial version. One of my concerns is that I am not 100% sure if my current implementation is good. Specifically, I am not sure if it will negatively impact the performance of the RL training. If there is someone who knows tensors well, I would like them to review it. However, this is not urgent.

**3. Default Radiative Properties**
As mentioned in point 1, my next target is to add exterior radiative heat transfer and codes for fenestrations. Thanks for sharing the deep research results. I would like to have something similar, but before setting the default values, I would like to first define what kinds of building materials we need to include. Based on my conversation with John, we don't need to define materials. Instead, we can classify them as:
* exterior wall
* interior wall
* outdoor air
* glazings (fenestration or transparent walls)

Once the classification is well-established, I can set the default values (and the default values will likely be different for each of the four cases).

So, let me add some more features for this PR. 
During the time, if you have some updates/information regarding the classification/codings, it would be great. (or we can define it in somewhere in docs if there is nothing well established yet.)

s2t2:
Updates from our discussion...

1. Looks like there are different values used in the dataset vs the building floorplan. And since these values in question are temporary (used for calculation purposes) and not a part of the persisted floorplan, we can leave as is. If we do need to add any specific numbers, we should make sure to add them to "simulator/constants.py".

2. Next steps for the tensorized approach include:

   + Comparing the results of the iterative vs tensorized approach to make sure they reconcile.
   + Maybe also observing some benchmark times (but we can even do this later). Not required at this time.

3. For the default radiation property values, let's make some new child classes with different defaults for the different types of materials. Like this:

```
class DefaultInsideAirRadiationProperties(RadiationProperties):
  alpha: float = 0.0  # absorptivity
  epsilon: float = 0.0  # emissivity
  tau: float = 1.0  # transmittance
  
class DefaultInsideWallRadiationProperties(RadiationProperties):
  alpha: float = 0.9  # absorptivity
  epsilon: float = 0.9  # emissivity
  tau: float = 0.0  # transmittance

class DefaultExteriorWallRadiationProperties(RadiationProperties):
  alpha: float = 0.9  # absorptivity
  epsilon: float = 0.9  # emissivity
  tau: float = 0.0  # transmittance
```

Then in the building, in `_assign_radiative_heat_transfer_properties`, we can use these defaults if none are passed in as params:

```
      inside_wall_radiative_properties = (
          inside_wall_radiative_properties
          or DefaultInsideWallRadiationProperties()
      )
      building_exterior_radiative_properties = (
          building_exterior_radiative_properties
          or DefaultExteriorWallRadiationProperties()
      )
      inside_air_radiative_properties = (
          inside_air_radiative_properties
          or DefaultInsideAirRadiationProperties()
      )
```

ecosang:
@s2t2 I added 
(1) validation case for iterative approach vs tensorized version approach,
(2) default values for radiative heat transfer properties.
For (2), I am not sure if the way I define the base class and default classes is correct.  https://github.com/google/sbsim/pull/94/commits/33dbb0328470df207f6427b45b6896931f163a1b

Otherwise, I think the interior radiative heat transfer is good to go merging. 
I will create a separate PR for exterior wall radiant heat transfer.

s2t2:
@ecosang ok thanks! 

one question before we merge. I see you recently added this validation logic to ensure that the sum of a material's emissivity + absorptivity + transmittance should either equal 0 or 1. is that right? can you explain this some more please? I am getting some conflicting responses from Gemini about whether this is valid. 


https://github.com/google/sbsim/pull/94/files#diff-8da221bbfb0189be8650282ce9ff91ae68818902fe6d72a08cb635b27a2fb74aR92-R97


```
    # Check that the sum of radiative properties is either 0 or 1
    total = self.alpha + self.epsilon + self.tau
    if not (abs(total - 0.0) < 1e-10 or abs(total - 1.0) < 1e-10):
      raise ValueError(
          f"The sum of alpha ({self.alpha}), epsilon ({self.epsilon}), "
          f"and tau ({self.tau}) must be either 0 or 1, but got {total}."
      )
```

I did read something related to absorptivity, *reflectivity*, and transmittance summing to one, however: 

> Conservation of Energy: When radiant energy strikes a surface, it can be absorbed, reflected, or transmitted. The Law of Conservation of Energy requires that the fractions of energy undergoing these processes must sum to 1. Letting $\rho$ be the reflectivity:
$\alpha + \rho + \tau = 1$
This means the sum of absorptivity, reflectivity, and transmittance is always 1.
>
> Kirchhoff's Law of Thermal Radiation: For a material in thermodynamic equilibrium with its surroundings, its emissivity ($\epsilon$) is equal to its absorptivity ($\alpha$):
$\epsilon = \alpha$
This holds true for any given wavelength and direction.
>
> The sum $\alpha + \epsilon + \tau$ does NOT generally equal 1 or 0. As shown above, the properties that sum to 1 are absorptivity, reflectivity, and transmittance. Since $\epsilon = \alpha$ (at equilibrium), there's no physical law stating $\alpha + \epsilon + \tau$ must be 1.

ecosang:
@s2t2 
Gosh.. Thanks for the catch. While I was reviewing my old textbooks, I totally messed up some concepts and made mistakes.
You are right. alpha+tau+rho=1 while epsilon is independent. 
For the interior radiative heat transfer, only `epsilon` is used at this point. 
I fixed and updated the default values based on other source.

But for the test cases, I maintained the original my default values instead of the new default values. But this is just a test case, I think it is fine.. I can update it, but I need to update all the numeric values for testing cases, which takes time, and if we update the default values later, it will break the test case all the time.

s2t2:
@ecosang OK thanks! Good catch 😸 

I was able to get the data class inheritance working (the issue had to deal with default values).

I also updated the formatting of the table in the docstring to ensure it shows up in the docs site.

I have that in a local commit, and will push soon. Just need to rebase first to include your latest commit and resolve any conflicts.

s2t2:
@ecosang Follow-up question: Is there any scenario where the sum of alpha + rho + tau equals zero (instead of one)? And what would this mean? It is my understanding that the sum needs to equal one always, but please let me know your thoughts. I am happy to update the validation logic to require the sum to equal one.

ecosang:
@s2t2 yes. It is always 1. If you can update, it will be great.  I am kind of having 0 option just in case to put all 0 for some reason, but it seems like it is not necessary as we have an option to exclude the radiative heat transfer calculation.

s2t2:
OK I just updated the tests to use the new default values.

I also updated the test data accordingly. I started saving the test data in the repo, to make the tests more readable, and to allow us more lasting access to the resulting values (which will help us troubleshoot and diagnose in the future as necessary). 

I also slightly refactored / reorganized some of the test files, which hopefully improves maintainability and modularity.

s2t2:
@ecosang I just re-reviewed and things are looking good from my perspective.

Could you give one final review, make any final updates you would like, and let me know once approved. Then we will merge! 🚀  

One comment for consideration is that I created a shared helper function called `create_building_with_radiative_properties` in the "simulator/conftest.py" that can be used to create a `FloorPlanBasedBuilding` with the default radiation props. I saw there was a similar function `_create_test_building_radiative` in the "simulator/tf_simulator_test.py" and also related code in the `_create_small_building` in the "simulator_flexible_floorplan_test.py" - perhaps there may be an opportunity to leverage the shared helper function in these spots.

s2t2 (approved):

ecosang:
@s2t2 thanks Michael. That's awesome. Thanks for organizing and making more testing cases. Looks good to me.!

s2t2:
@ecosang sorry about that table getting messed up / lost. 

Here is the commit where I formatted it and made sure it was visible in the docs site:
https://github.com/google/sbsim/pull/94/commits/478fbe45f4c2d169fde47b7dc4e35ecaf62dcc92#diff-8da221bbfb0189be8650282ce9ff91ae68818902fe6d72a08cb635b27a2fb74aR79-R91

Looks like it got removed with this commit:
https://github.com/google/sbsim/pull/94/commits/4d7e75d494b056fa61b9860d1609ea771563f78b

If you would like to submit a quick documentation fix in a separate PR, we can prioritize merging that right away.

PR #97, opened 2025-06-09
Dataset Tests and Refactoring

s2t2:
Tests and refactors the `SmartBuildingsDataset` class currently defined in "smart_control/dataset/dataset.py", using an object oriented approach.

Details:

 + Refactors the dataset class into two separate classes: a `BuildingDataset` class to handle building-level data, and a `BuildingDatasetPartition` class to handle partition-specific data. 

  + Parameterizes the dataset building name. This enables future extensibility to other buildings besides "sb1".

  + Adds tests for the dataset download process, as well as to describe the structure of the resulting data. This includes  tests for dataset downloading. Skips these tests by default, because it takes around two minutes to download and unzip the data and we don't want to slow down the build. However we can run these tests as desired by passing a `TEST_DATASET_DOWNLOAD` environment variable. After the data is already downloaded, the tests will use existing data, so subsequent runs are much faster. The tests utilize a pytest session-level fixture to share the downloaded dataset across multiple test files.


smart_control/dataset/dataset.py
```diff
@@ -1,96 +1,355 @@
...
-      return local_filename
+    return local_filepath
 
-    url = "https://storage.googleapis.com/gresearch/smart_buildings_dataset/tabular_data/sb1.zip"  # pylint: disable=line-too-long
-    download_file(url)
-    shutil.unpack_archive("sb1.zip", "sb1/")
+  def download(self):
+    """Downloads the building's dataset from Google Cloud Storage."""
+    self._download_file_if_not_exists(self.zip_url)
+    shutil.unpack_archive(self.zip_filepath, self.building_dirpath)
```
s2t2: Move this inside the download if not exists function, to prevent unnecessary unpacking/ unzipping.

smart_control/dataset/dataset.py
```diff
@@ -1,96 +1,355 @@
...
 
-  def get_floorplan(self, building):
-    """Gets the floorplan and device layout map for a specific building.
+  @property
+  def floorplan_image_filepath(self):
+    floorplan_image_filename = f"{self.building_id}_floorplan.png"
+    return os.path.join(self.building_dirpath, floorplan_image_filename)
+
+  def display_floorplan(
+      self, cmap="binary", show=True, save=True, image_filepath=None
```
s2t2: Use a string type hint for the image filepath

smart_control/dataset/dataset.py
```diff
@@ -1,96 +1,355 @@
...
+    ![An image of a floorplan.](../assets/images/sb1_floorplan.png)
 
     Args:
-      building: The name of the building.
+      cmap (str): The name of a [matplotlib color map](https://matplotlib.org/stable/users/explain/colors/colormaps.html)
+        to use when rendering the image.
+      show (bool): Whether or not to show the image.
+      save (bool): Whether or not to save the image (as a .png file).
+      image_filepath (str): A custom filepath to use when saving the image.
+        Only applies if `save=True`.
```
s2t2: Add a note about the default image save location

smart_control/dataset/dataset_test.py
```diff
@@ -0,0 +1,524 @@
...
+
+  if os.path.exists(ZIP_FILEPATH):
+    os.remove(ZIP_FILEPATH)
+
+  if os.path.exists(DATASET_DIRPATH):
+    shutil.rmtree(DATASET_DIRPATH)
+
+
+class TestDataDirectory(absltest.TestCase):
+  'Tests for the data directory.'
```
s2t2: Use triple quotes

smart_control/dataset/dataset_test.py
```diff
@@ -0,0 +1,524 @@
...
+    ]
+    self.assertEqual(device_keys, expected_device_keys)
+
+    # each device layout map is a list of two integer values:
+    # the layout shapes are not the same across all devices
+    first_layout_map = device_layout_map['VAV CO 1-1-06']
+    self.assertIsInstance(first_layout_map, list)
+    self.assertEqual(np.array(first_layout_map).shape, (1021, 2))
+    self.assertEqual(first_layout_map[0], [79, 35])
+    self.assertEqual(first_layout_map[-1], [80, 64])
```
s2t2: Also Test the last layout map to show it has a different length

smart_control/dataset/dataset_test.py
```diff
@@ -0,0 +1,524 @@
...
+
+  @unittest.skipUnless(TEST_DATASET_DOWNLOAD, SKIP_REASON)
+  def test_device_infos(self):
+    device_infos = self.ds.device_infos
+    self.assertIsInstance(device_infos, list)
+    self.assertEqual(len(device_infos), 173)
+
+    # first item / example structure:
+    first_device_info = {
```
s2t2: If this structure is shared with the df test, move it to a share place, probably a property on the test class.

s2t2: Oh this one has the zone id whereas that has been removed from the df

smart_control/dataset/dataset_test.py
```diff
@@ -0,0 +1,524 @@
...
+    self.assertEqual(devices_df.iloc[0].to_dict(), first_row)
+
+  @unittest.skipUnless(TEST_DATASET_DOWNLOAD, SKIP_REASON)
+  def test_zone_infos(self):
+    zone_infos = self.ds.zone_infos
+
+    self.assertIsInstance(zone_infos, list)
+    self.assertEqual(len(zone_infos), 563)
+
+    first_zone_info = {
```
s2t2: Consider moving to a shared location like a property on the test class that can also be accessed by the df test

smart_control/dataset/dataset_test.py
```diff
@@ -0,0 +1,524 @@
...
+
+  @classmethod
+  def setUpClass(cls):
+    super().setUpClass()
+    cls.partition = BuildingDatasetPartition(
+        building_dataset=cls.ds, partition_id='2022_a'
+    )
+
+  @classmethod
+  def tearDownClass(cls):
```
s2t2: This child class shouldn't need the tear down

smart_control/dataset/dataset.py
```diff
@@ -1,96 +1,469 @@
...
+
+  @cached_property
+  def metadata(self) -> dict:
+    """Metadata describing the partition [data](./#smart_control.dataset.dataset.BuildingDatasetPartition.data).
+
+    Returns:
+      A dictionary containing the following keys:
+
+        - `'action_ids'`: dictionary mapping action string identifiers to
+          integer indices, for example: `{'action_id@setpoint': 0}`
```
s2t2: Todo: verify the meaning of the key names for `action_ids`, `observation_ids`, and `reward_ids`

smart_control/dataset/dataset.py
```diff
@@ -1,96 +1,469 @@
...
+        - `'observation_timestamps'`: list of sequential of timestamps,
+          representing the time of each observation
+
+        - `'reward_ids'`: dictionary mapping reward string identifiers to
+          integer indices, for example: `{'reward_id@type': 0}`
+
+        - `'reward_timestamps'`: list of sequential timestamps, representing the
+          time of each reward
+
+        - `'reward_info_timestamps'`: list of sequential timestamps, related to
```
s2t2: question: what is the difference between between (`reward_timestamps` / `reward_value_matrix`) and (`reward_info_timestamps` / `reward_info_value_matrix`)? looks like they both use the `reward_ids`. 

@JudahGoldfeder

smart_control/dataset/dataset.py
s2t2: I think the `action_ids` and the `observation_ids` are both in the format of `device_id@setting_name`. But not sure about the `reward_ids` - it seems the first value is either the `zone_id` or the `device_id`.

s2t2:
superseded by #101

Issue #34, opened 2025-05-01
Improve Test Coverage


s2t2:
We should aim to test all public-facing functions. Among other benefits, this will allow us to refactor the code with more confidence, without worrying we are breaking things.

This issue can serve as an "epic". As collaborators add tests for previously untested code, we can tag this issue to show progress towards this goal. 

When we have satisfactory test coverage, we can consider closing this issue.

PR #101, opened 2025-06-18
Dataset Tests and Refactoring

s2t2:
Refactors the `SmartBuildingsDataset` class currently defined in "smart_control/dataset/dataset.py", using an object oriented approach.

Details:

 + Refactors the dataset class into two separate classes: a `BuildingDataset` class to handle building-level data, and a `BuildingDatasetPartition` class to handle partition-specific data. 

  + Parameterizes the dataset identifier. This enables future extensibility to other buildings besides "sb1".

  + Adds tests for the dataset download process, as well as to describe the structure of the resulting data. This includes  tests for dataset downloading. Skips these tests by default, because it takes around two minutes to download and unzip the data and we don't want to slow down the build. However we can run these tests as desired by passing a `TEST_DATASET_DOWNLOAD` environment variable. After the data is already downloaded, the tests will use existing data, so subsequent runs are much faster. The tests utilize a pytest session-level fixture to share the downloaded dataset across multiple test files. #34 

+ Adds documentation for the new classes. #105 

+ Adds functionality to display the building floorplan as an image, and to save the image to disk. Displays the floorplan image in the docs site.

+ Adds functionality to construct pandas dataframe objects from building and partition data. Displays example dataframes in the docs.

---

### Migration Guide

This PR makes some breaking changes to the dataset interface. Here is a migration guide for anyone who was using the old interface...

```py
# old:
ds = SmartBuildingDataset()
floorplan, device_layout_map = ds.get_floorplan("sb1")

# new:
ds = BuildingDataset("sb1")
floorplan = ds.floorplan
device_layout_map = ds.device_layout_map
```

```py
# old:
ds = SmartBuildingDataset()
data, metadata = ds.get_building_data("sb1", "2022a")

# new:
ds = BuildingDataset("sb1")
partition = BuildingDatasetPartition("2022a")
data = partition.data
metadata = partition.metadata
```

NOTE: we were previously adding the building's zone info and device info into the partition's metadata, however now we are preferring to reach into the dataset for this information.


smart_control/dataset/dataset.py
```diff
@@ -1,96 +1,405 @@
...
-    return data, metadata
+    # pylint: enable=line-too-long
+    df = pd.DataFrame(self.device_infos)
+    df = df.drop(columns=["zone_id"])  # many to many relationship
+    # consider renaming zone_id to zone_ids instead of dropping
+    return df
+
+  # DEVICE FIELDS
+
+  def _count_device_fields(self, field_type: str) -> pd.Series:
```
s2t2, line 275: If we value efficiency and avoiding the nested loop, we could consider using a [`polars`](https://pola.rs/) Dataframe and/or SQL Context to write a query to produce the same results. However if we value keeping the package dependencies to a minimum, we can move forward with the current approach, and circle back to improving efficiency in the future as needed.

smart_control/dataset/dataset.py
```diff
@@ -1,96 +1,405 @@
...
+VALID_BUILDING_PARTITIONS = {
+    "sb1": ["2022_a", "2022_b", "2023_a", "2023_b", "2024_a"]
+}
+
+
+class BuildingDataset:
+  """A helper class for handling the dataset for a specific building.
+
+  Args:
+    building_id (str): The identifier of the building (e.g. "sb1").
```
s2t2: Here the `building_id` is "sb1" (a short code uniquely identifying the building dataset). However we see the zone information contains a different value / scheme for the `building_id` (like 'buildings/3616672508'). 

In light of this, we might consider renaming the `building_id` parameter to something like `building_code` , `building_name`, `building_dataset_id`, or `dataset_id`.

smart_control/dataset/dataset_partition.py
```diff
@@ -0,0 +1,420 @@
...
+    return list(self.observation_ids_map.keys())
+
+  @cached_property
+  def reward_ids(self) -> list[str]:
+    """A list of unique reward identifiers.
+
+    Reward identifiers are in the format of `device_id@field_name` or
+    `zone_id@field_name`.
+    For example: `'rooms/9028552126@heating_setpoint_temperature'` or
+    `'14409954889734029312@air_conditioning_electrical_energy_rate'`.
```
s2t2: It would be nice to confirm the format of the reward identifiers.

smart_control/dataset/dataset_partition.py
```diff
@@ -0,0 +1,420 @@
...
+    return self.metadata["observation_timestamps"]
+
+  @cached_property
+  def reward_timestamps(self) -> list[pd.Timestamp]:
+    """A list of sequential timestamps representing the time of each reward."""
+    return self.metadata["reward_timestamps"]
+
+  @cached_property
+  def reward_info_timestamps(self) -> list[pd.Timestamp]:
+    """A list of sequential timestamps related to reward information."""
```
s2t2: It would be nice to confirm the difference between rewards and reward infos.

smart_control/dataset/dataset_partition_test.py
```diff
@@ -0,0 +1,468 @@
...
+
+  @unittest.skipUnless(TEST_DATASET, SKIP_REASON)
+  def test_observation_value_matrix(self):
+    self.assertEqual(self.partition.observation_value_matrix.shape, (51852, 1198))  # pylint: disable=line-too-long
+
+  @unittest.skipUnless(TEST_DATASET, SKIP_REASON)
+  def test_reward_value_matrix(self):
+    # QUESTION: there are 3252 reward identifiers, but only 17 columns here
+    # so what are these 17 columns about?
+    self.assertEqual(self.partition.reward_value_matrix.shape, (51852, 17))
```
s2t2, line 247: QUESTION / ISSUE: there are 3252 reward identifiers, and 3252 columns in the `reward_info_value_matrix`, but only 17 columns in the `reward_value_matrix`. So what are these 17 columns about? Or is there an issue with the dataset?

@JudahGoldfeder

smart_control/dataset/dataset_partition.py
s2t2: Comes from the Protos. Four Protos. Reward Info is the set of raw information used to calculate rewards (raw inputs to the reward function, which produces a single scalar value). Reward response contains ?

smart_control/dataset/dataset_partition_test.py
s2t2: The 17 values correspond with the Reward response proto (not the timestamps)

https://github.com/google/sbsim/blob/copybara_push/smart_control/proto/smart_control_reward.proto#L124

smart_control/dataset/dataset_test.py
```diff
@@ -0,0 +1,498 @@
...
+from smart_control.dataset.conftest import TEST_DATASET
+from smart_control.dataset.conftest import ZIP_FILEPATH
+from smart_control.dataset.dataset import BuildingDataset
+from smart_control.dataset.dataset import DATA_DIR
+
+#
+# EXAMPLE BUILDING LEVEL DATA
+#
+
+_BUILDING_ID = 'sb1'
```
s2t2: Update var name

smart_control/dataset/dataset_test.py
```diff
@@ -0,0 +1,498 @@
...
+    self.assertEqual(floorplan.shape, (744, 1004))
+
+    values, counts = np.unique(floorplan, return_counts=True)
+    value_counts = dict(zip(values, counts))
+    self.assertEqual(value_counts, {0.0: 436332, 1.0: 60204, 2.0: 250440})
+
+  @unittest.skipUnless(TEST_DATASET, SKIP_REASON)
+  def test_display_floorplan(self):
+    # setup:
+    filepath = os.path.join(os.path.dirname(__file__), 'tmp_floorplan.png')
```
s2t2: Use tempfile

smart_control/dataset/dataset_test.py
```diff
@@ -0,0 +1,498 @@
...
+    'supply_water_temperature_sensor',
+    'supply_water_temperature_setpoint',
+    'zone_air_co2_concentration_sensor',
+    'zone_air_co2_concentration_setpoint',
+    'zone_air_cooling_temperature_setpoint',
+    'zone_air_heating_temperature_setpoint',
+    'zone_air_temperature_sensor',
+]
+
+_DEVICE_ACTIONABLE_FIELD_NAMES = [
```
s2t2, line 224: Consider moving to conftest

smart_control/dataset/dataset_test.py
```diff
@@ -0,0 +1,498 @@
...
+    observable_fields = self.ds.observable_fields
+    self.assertIsInstance(observable_fields, list)
+    self.assertEqual(len(observable_fields), 51)
+    self.assertEqual(observable_fields, _DEVICE_OBSERVABLE_FIELD_NAMES)
+
+    value_counts = self.ds.observable_field_counts
+    self.assertIsInstance(value_counts, pd.Series)
+    self.assertEqual(len(value_counts), 51)
+    self.assertEqual(
+        value_counts.head(1).to_dict(),
```
s2t2: Do top 5

smart_control/dataset/dataset_test.py
```diff
@@ -0,0 +1,498 @@
...
+    self.assertEqual(len(actionable_fields), 27)
+    self.assertEqual(actionable_fields, _DEVICE_ACTIONABLE_FIELD_NAMES)
+
+    value_counts = self.ds.observable_field_counts
+    self.assertIsInstance(value_counts, pd.Series)
+    self.assertEqual(len(value_counts), 51)
+    self.assertEqual(
+        value_counts.head(1).to_dict(),
+        {'supply_air_damper_percentage_command': 123},
+    )
```
s2t2, line 433: Consider top 5 bottom 5

smart_control/dataset/dataset_partition_test.py
```diff
@@ -0,0 +1,500 @@
...
+    reward_infos_shape=(52140, 3252),
+    reward_infos_range=('2023-12-31 23:55:00+00:00', '2024-06-30 00:50:00+00:00'),
+  ),
+]
+# pylint:enable=line-too-long
+# fmt: on
+
+
+@pytest.mark.usefixtures('set_dataset')
+class TestAllBuildingDatasetPartitions(parameterized.TestCase):
```
s2t2, line 83: Consider renaming the partition files to partition.py and partition_test.py

smart_control/dataset/dataset_partition_test.py
```diff
@@ -0,0 +1,500 @@
...
+from smart_control.dataset.conftest import TEST_DATASET
+from smart_control.dataset.dataset_partition import BuildingDatasetPartition
+
+#
+# HIGH LEVEL TESTS FOR ALL PARTITIONS IN BUILDING "SB1"...
+#
+
+# pylint:disable=line-too-long
+# fmt:off
+PARTITION_PARAMETERS = [
```
s2t2, line 21: Move closer to test

smart_control/dataset/dataset_partition_test.py
```diff
@@ -0,0 +1,500 @@
...
+    self.assertEqual(df.dtypes.unique().tolist(), [np.dtype('float64')])
+
+  @unittest.skipUnless(TEST_DATASET, SKIP_REASON)
+  def test_reward_infos_df(self):
+    df = self.partition.reward_infos_df
+
+    self.assertIsInstance(df, pd.DataFrame)
+    self.assertEqual(df.shape, (51852, 3252))
+
+    # columns are the reward ids:
```
s2t2: Reward info ids

smart_control/dataset/dataset_partition.py
```diff
@@ -0,0 +1,496 @@
...
+        "action_ids_map": metadata["action_ids"],  # renamed
+        "action_timestamps": metadata["action_timestamps"],
+        "observation_ids_map": metadata["observation_ids"],  # renamed
+        "observation_timestamps": metadata["observation_timestamps"],
+        "reward_info_ids_map": metadata["reward_ids"],  # renamed
+        "reward_timestamps": metadata["reward_timestamps"],
+        "reward_info_timestamps": metadata["reward_info_timestamps"],
+    }
+
+    # here we were previously redundantly adding information from the dataset:
```
s2t2: Move comment to change log

smart_control/dataset/dataset.py
```diff
@@ -1,96 +1,406 @@
...
+  def floorplan_filepath(self):
+    return os.path.join(self.tabular_dirpath, "floorplan.npy")
 
-  def get_floorplan(self, building):
-    """Gets the floorplan and device layout map for a specific building.
+  @cached_property
+  def floorplan(self) -> np.ndarray:
+    """The building's floorplan, as a numpy array.
+
+    The floorplan consists of a map of pixels. Here is a mapping of the values:
```
s2t2: "With the following values"

s2t2:
All dataset tests passing locally (most of these are skipped on the CI build):

Screenshot 2025-07-10 2 12 03 PM

PR #109, opened 2025-07-26
fix: Replaced assert with ValueError in stochastic_occupancy

### LINKED ISSUE #28: Remove assertions from non-test code
Let's remove all the assertions from non-test code. For example the ones occurring in `smart_control/simulator/stochastic_occupancy.py`.

> "Do not use assert statements in place of conditionals or validating preconditions." - [source](https://google.github.io/styleguide/pyguide.html#24-exceptions)

_Originally posted by @s2t2 in https://github.com/google/sbsim/pull/27#discussion_r2054799532_

Instead, we should raise errors like `ValueError`,  etc.

igopalakrishna:
Fixes #28

This pull request resolves the issue by replacing `assert` statements with explicit `ValueError` exceptions in the example file mentioned, `smart_control/simulator/stochastic_occupancy.py`.

Using `ValueError` is better practice as `assert` statements can be disabled in production environments, which would otherwise silently ignore important precondition checks.

- [x] Tests pass
- [x] No documentation changes were necessary for this change.

### DIFF
--- a/smart_control/simulator/stochastic_occupancy.py
+++ b/smart_control/simulator/stochastic_occupancy.py
@@ -54,13 +54,22 @@ class ZoneOccupant:
       random_state: np.random.RandomState,
       time_zone: Union[datetime.tzinfo, str] = "UTC",
   ):
-    assert (
+    # Validate that the time bounds are in chronological order
+    if not (
         earliest_expected_arrival_hour
         < latest_expected_arrival_hour
         < earliest_expected_departure_hour
         < latest_expected_departure_hour
-    )
-    assert lunch_start_hour < lunch_end_hour
+    ):
+      raise ValueError(
+          "Time bounds must be in chronological order i.e., "
+          "expected arrival/departure hours must be strictly increasing:" "earliest_expected_arrival_hour < latest_expected_arrival_hour "
+          "< earliest_expected_departure_hour < latest_expected_departure_hour."
+      )
+
+    # Validate lunch time bounds
+    if lunch_start_hour >= lunch_end_hour:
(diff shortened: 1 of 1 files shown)


smart_control/simulator/stochastic_occupancy.py
```diff
@@ -54,13 +54,22 @@ def __init__(
...
         earliest_expected_arrival_hour
         < latest_expected_arrival_hour
         < earliest_expected_departure_hour
         < latest_expected_departure_hour
-    )
-    assert lunch_start_hour < lunch_end_hour
+    ):
+      raise ValueError(
+          "Time bounds must be in chronological order i.e., "
+          "expected arrival/departure hours must be strictly increasing:" "earliest_expected_arrival_hour < latest_expected_arrival_hour "
```
s2t2, line 66: I think we want the second string on its own line, to avoid line-too-long errors.

s2t2: let's consider using more natural language in the error message, like "earliest expected arrival must come before latest expected arrival, earliest expected departure must come before latest expected departure", etc.

s2t2 (commented):
Nice, thanks @igopalakrishna for taking up this issue. 🙌  I provided some specific comments below for the first instance you fixed.

In general, it would be nice to have a single PR that addresses all the instances across the codebase.

With the error raising implemented, we should now be able to add some tests to ensure each of the errors is triggered when we pass in certain invalid conditions. If you are able to add a corresponding test for each error, that would be awesome!

s2t2:
@igopalakrishna should we close this PR in favor of #113 ?

s2t2:
I believe this PR is superseded and closed by #113

PR #111, opened 2025-07-27
feat: Optimize Dockerfile and update setup documentation

### LINKED ISSUE #80: Optimize Dockerfile
We introduced a Dockerfile in #77, however there may be opportunities to improve it. See comments in #77, including...

A) Simplifying installation of Python dependencies:

> We are using Pip to install some dependencies, and then later using Poetry to install additional dependencies. However it would be nice to consolidate the installation approach to rely on Poetry as much as possible. For example, it looks like Poetry is already installing `dm-reverb`, so we could perhaps remove that from the Pip installation step. If the rest of the dependencies are required for running notebooks, let's consider adding `jupyter` and any other notebook-related dependencies to the dev dependencies group (or a new "notebooks" group) in the "pyproject.toml" file.

B) Leveraging Dockerfile conventions:

> Running the Dockerfile contents through https://hadolint.github.io/hadolint/ provides the following recommendations for us to consider as well:
>
>  + Pin versions in apt get install. Instead of `apt-get install <package>` use `apt-get install <package>=<version>` -- [more info](https://github.com/hadolint/hadolint/wiki/DL3008)
>  + Use WORKDIR to switch to a directory, instead of cd -- [more info](https://github.com/hadolint/hadolint/wiki/DL3003)
>  + Set the SHELL option -o pipefail before RUN with a pipe in it -- [more info](https://github.com/hadolint/hadolint/wiki/DL4006)
>  + This ln has no destination. Check the arguments, or specify '.' explicitly. -- [more info](https://github.com/koalaman/shellcheck/wiki/SC2226)
>  + Either use Wget or Curl but not both -- [more info](https://github.com/hadolint/hadolint/wiki/DL4001)
>  + echo may not expand escape sequences. Use printf. -- [more info](https://github.com/koalaman/shellcheck/wiki/SC2028)

C) Choosing appropriate size and split of RUN commands for Linux installations:

> Gemini seems to advocate for splitting up the large RUN command for all the Linux installations, but ChatGPT seems to prefer to keep the single large RUN command. 

Instead of taking over the terminal, we would like to run the docker environment in a detached way so we can execute various commands against it, like running tests, running scripts, and running Jupyter notebooks as desired. 

We would also like to expand the Docker setup instructions.

This would be a great issue for anyone who is particularly familiar with Docker! 🐳

igopalakrishna:
Closes #80

This PR completely revamps the project's Docker setup to improve dependency management, build efficiency, and user experience, addressing all points outlined in the issue.

### A) Dependency Management (`pyproject.toml` & `Dockerfile`)
- Consolidated all Python package installation to use **Poetry**, removing the separate `pip install` step from the `Dockerfile`.
- Moved notebook-specific packages (`jupyter`, `ipykernel`, `ipython`) into a new `[tool.poetry.group.notebooks]` group in `pyproject.toml`.
- Added `grpcio-tools` as a development dependency to ensure reliable `.proto` file compilation within the Python environment.
- Updated the `protobuf` version to `^4.25.0` to resolve a dependency conflict with `grpcio-tools`.

### B) & C) Dockerfile Optimization
- **Best Practices:** The `Dockerfile` now follows modern conventions, including:
  - Using `WORKDIR` instead of `RUN cd`.
  - Setting `SHELL ["/bin/bash", "-o", "pipefail", "-c"]` to ensure build steps fail correctly.
  - Using `curl` consistently instead of `wget`.
- **Build Efficiency:**
  - `RUN` commands have been restructured into logical layers to improve Docker's layer caching and significantly speed up subsequent builds.
  - The build process now copies only `pyproject.toml` and `poetry.lock` before installing dependencies, preventing re-installation on minor code changes.
- **Simplicity:**
  - The complex startup script (`start.sh`) and the `git clone` step have been removed in favor of a cleaner `COPY . .` and a direct `CMD` instruction.
  - The Jupyter token is now disabled in the `CMD` for a better developer experience, matching the original `Dockerfile`'s intent.

### Expanded Documentation (`docs/setup/docker.md`)
- The Docker setup guide has been completely rewritten to be clearer and more comprehensive.
- The new guide provides instructions for running the container in both foreground and detached modes.
- It now includes specific commands for executing tasks against a running container, such as running the `pytest` suite or accessing an interactive shell, as requested in the issue.

### How to Test
1. Build the image: `docker build -t sbsim .`
2. Run the container: `docker run -p 8888:8888 sbsim`
3. Access Jupyter at `http://localhost:8888`.
4. Run tests against the detached container: `docker exec <container_name> poetry run pytest`

### Known Issue
The test suite fails with a `Fatal Python error: Illegal instruction` when run inside the container on an Apple Silicon (ARM) Mac. This is a known issue related to CPU instruction sets when running TensorFlow in an emulated environment and is outside the scope of the Dockerfile optimizations. The `Dockerfile` builds successfully, the Jupyter server runs, and the `docker exec` command is correctly configured for users on native x86/amd64 systems.


ajkhabbazi:
@igopalakrishna @s2t2 Tried testing this PR locally (Windows + Docker Desktop). The Docker build fails at the apt install step:

E: Unable to locate package python3.10-venv
E: Unable to locate package python3.10-distutils
(exit code 100)

CI also shows a failing ubuntu-latest (Py3.11) job. Suggest switching the base image to Ubuntu 22.04 (jammy) or adjusting the Python install approach, then re-running CI.

s2t2:
Thanks @ajkhabbazi for the review!

Let's use Python 3.11 in the Docker file, to match the other operating environments.

s2t2:
Let's revert the changes to the pb2.py files, and then as long as you think this is ready, we can merge!

ajkhabbazi:
@igopalakrishna -- two fixes needed based on @s2t2's feedback:
 1. Switch base image to ubuntu:22.04 and replace all python3.10 references with python3.11 -- fixes the apt install failure on Ubuntu 20.04.
 2. Revert the three pb2.py files -- they're auto-generated and shouldn't be in this PR.
Happy to take over and complete this PR if you're no longer working on it.

s2t2:
@ajkhabbazi if you want to get this past the finish line, that would be great!

ajkhabbazi:
@s2t2 Done -- opened #173.

Carries @igopalakrishna's commits forward with authorship intact. On top: merged `copybara_push` (10 commits behind, only `poetry.lock` conflicted -- regenerated it), and reverted the three `*_pb2.py` files per your April note.

Verified in the container: builds clean, pytest passes (830/37 skipped), and the `*_pb2.py` files regenerate at build time.

Close this one once #173 merges.

PR #113, opened 2025-07-27
Replace Non-test Assertions with Value Errors

### LINKED ISSUE #28: Remove assertions from non-test code

igopalakrishna:
This PR addresses issue #28 by systematically replacing `assert` statements with explicit exception raising (primarily `ValueError`) in non-test production code within the `smart_control` directory.

`assert` statements can be optimized away in Python when run with the `-O` flag, leading to silent failures. Replacing them with explicit `raise` statements ensures that critical preconditions and validation checks are always enforced, improving the robustness and reliability of the application.

**Key Changes and Files Modified:**

* **`smart_control/environment/environment.py`**:
    * Replaced `assert` statements with `ValueError` checks in `_get_observation_spec_histogram_reducer` and `_commit_reward_metrics` to ensure proper configuration of internal components.
* **`smart_control/reward/natural_gas_energy_cost.py`**:
    * Replaced `assert` with `ValueError` in `__init__` to validate that `gas_price_by_month` has exactly 12 values.
* **`smart_control/reward/setpoint_energy_carbon_regret.py`**:
    * Replaced `assert` with `ValueError` in `__init__` to ensure `max_productivity_personhour_usd` is greater than `min_productivity_personhour_usd`.
* **`smart_control/simulator/boiler.py`**:
    * Replaced `assert` with `ValueError` in `compute_thermal_dissipation_rate` to validate that `water_temp` is not less than `outside_temp`, including values in the error message for debugging.
* **`smart_control/simulator/randomized_arrival_departure_occupancy.py`**:
    * Replaced `assert` with `ValueError` in `ZoneOccupant.__init__` and `_get_event_probability` to ensure chronological order and validity of time bounds, including values in error messages.
* **`smart_control/simulator/simulator.py`**:
    * Replaced `assert` statements with `ValueError` checks in `_get_corner_cv_temp_estimate`, `_get_edge_cv_temp_estimate`, and `_get_interior_cv_temp_estimate` to validate the number of neighbors for different Control Volume types, including neighbor lists for debugging.
* **`smart_control/simulator/stochastic_occupancy.py`**:
    * Replaced `assert` with `ValueError` in `ZoneOccupant.__init__` to validate chronological order of time bounds and the lunch break hours.
* **`smart_control/simulator/vav.py`**:
    * Replaced `assert` with `ValueError` in `max_air_flow_rate` setter.
    * Replaced two `assert`s with a single, combined `ValueError` check in `compute_zone_supply_temp` to ensure the calculated `air_flow_rate` is positive, preventing division by zero.
* **`smart_control/utils/controller_reader.py`**:
    * Replaced `assert` with `ValueError` in `ProtoReader._select_shards._read_timestamp`.
    * **_Added a critical robustness improvement_**: Implemented an additional check to prevent `IndexError` if the timestamp regex finds no matches in the filepath, raising a more descriptive `ValueError`.
* **`smart_control/utils/conversion_utils.py`**:
    * Replaced `assert`s with `ValueError` in `normalize_dow` and `normalize_hod` to validate input ranges.
    * **_Significant improvement to `zone_id_to_coordinates`_**: Replaced the original regex with a stricter, anchored regex (`^...$`) and improved the error message to clearly state the expected format and the received invalid input (`{zone_id!r}`).
* **`smart_control/utils/energy_utils.py`**:
    * Replaced `assert`s with `ValueError` for sequence length validation in `get_humidity_ratio` and `get_air_conditioning_energy_rate`.
    * **_Crucial additional input validation_**: Added `for` loops with `ValueError` checks in `get_humidity_ratio` and `get_air_conditioning_energy_rate` to validate the *values* of `relative_humidities` (must be `(0,1]`) and `pressures` (must be `>0`), preventing physical impossibilities and potential `ZeroDivisionError` or `MathDomainError` downstream.
    * Refined the logic in `get_fan_power` to correctly handle `brake_hp` being `0` by changing `if brake_hp:` to `if brake_hp is not None:`.
* **Test Files (`*_test.py` files)**: Confirmed that `smart_control/observers/replay_buffer/replay_buffer_test.py` and `smart_control/utils/observation_normalizer_test.py` did not require changes, as `assert` statements are appropriate and expected within test suites.

This comprehensive set of changes ensures that the codebase provides more informative error messages and is more robust against invalid inputs and internal inconsistencies.

Closes #28


ajkhabbazi:
@igopalakrishna @s2t2 Tested locally (Windows + Docker Desktop) by checking out the PR branch and running `pytest -q` in the container. I’m seeing 3 test failures (669 passed, 37 skipped).

All failures are due to exception type: the updated code now raises `ValueError` (instead of `assert`), but the tests still expect `AssertionError`:
- `boiler_test.py`: expects `AssertionError`, but `boiler.py` raises `ValueError` when `water_temp < outside_temp`.
- `vav_test.py`: expects `AssertionError`, but `vav.py` raises `ValueError` when `air_flow_rate <= 0`.

Suggestion: update these tests to expect `ValueError` (or align on a single exception type) so CI passes.

igopalakrishna:
@ajkhabbazi Thanks a lot for testing and the detailed notes!

I updated the failing tests to expect ValueError where the implementation now raises ValueError (e.g., boiler_test.py and vav_test.py), and I also merged the latest upstream/copybara_push to resolve the previous conflict. The PR now shows no conflicts.

If you get a chance to re-run pytest, please let me know if anything else looks off.

ajkhabbazi:
@igopalakrishna @s2t2 Thanks for the update and for fixing the test issues. I pulled the latest version locally (Windows + Docker Desktop) and re-ran pytest -q. Everything passes on my side now (746 passed, 37 skipped, warnings only). LGTM.

smart_control/reward/natural_gas_energy_cost_test.py
```diff
@@ -69,6 +69,15 @@ def test_invalid_carbon_cost(self):
     energy_rate = -1.0
     self.assertEqual(0.0, cost.cost(start_time, end_time, energy_rate))
 
+  def test_invalid_gas_price_by_month_length(self):
```
s2t2, line 72: Nice, appreciate this test.

smart_control/reward/setpoint_energy_carbon_regret_test.py
```diff
@@ -280,6 +280,29 @@ def _get_test_reward_info(
 
     return info
 
+  def test_invalid_productivity_bounds(self):
+    """ValueError if max_productivity <= min_productivity."""
+    electricity_cost = TestEnergyCost(0.05, 0.01)
+    natural_gas_cost = TestEnergyCost(0.05, 0.01)
```
s2t2: let's prefer to use named parameters here, for improved readability

smart_control/simulator/randomized_arrival_departure_occupancy_test.py
```diff
@@ -141,6 +141,37 @@ def test_peek(self, tz):
...
+          latest_expected_departure_hour=18,
+          step_size=step_size,
+          random_state=random_state,
+      )
+
+  def test_get_event_probability_invalid_hours(self):
+    """ValueError when start_hour >= end_hour."""
+    random_state = np.random.RandomState(seed=55213)
+    step_size = pd.Timedelta(5, unit='minute')
+    occupant = randomized_arrival_departure_occupancy.ZoneOccupant(
```
s2t2, line 166: let's prefer to use named params here please

smart_control/utils/energy_utils.py
```diff
@@ -63,7 +63,21 @@ def get_humidity_ratio(
 
   Returns: water mass to air mass ratio in kg Water / kg Air
   """
-  assert len(temps) == len(relative_humidities) == len(pressures)
+  if len(temps) != len(relative_humidities) or len(temps) != len(pressures):
```
s2t2: let's prefer to use something like `if not (len(temps) == len(relative_humidities) == len(pressures))` here, to more closely match the original implementation

smart_control/utils/energy_utils.py
```diff
@@ -63,7 +63,21 @@ def get_humidity_ratio(
...
+    raise ValueError(
+        f'Input arrays must have equal length. Got: temps={len(temps)}, '
+        f'relative_humidities={len(relative_humidities)}, '
+        f'pressures={len(pressures)}.'
+    )
+
+  # Sanity-check each RH and pressure
```
s2t2: would it be possible to use numpy-based methods here, like `np.all` or `np.any`?

smart_control/utils/energy_utils.py
```diff
@@ -63,7 +63,21 @@ def get_humidity_ratio(
...
+  for i, rh in enumerate(relative_humidities):
+    if rh <= 0.0 or rh > 1.0:
+      raise ValueError(f'relative_humidities[{i}] must be in [0,1], got {rh}.')
```
s2t2: Gemini says:

> Note: Your original error message mentioned the range [0,1], but your code logic `rh <= 0.0` actually excludes zero. If you want to include zero, change the logic to `rh < 0.0`.

if we do want to exclude zero, perhaps we could change the message to `(0, 1]` which is exclusive of zero (whereas `[0,1]` is inclusive)

smart_control/utils/energy_utils_test.py
```diff
@@ -43,6 +72,19 @@ def test_get_air_conditioning_energy_rate(self):
...
     self.assertAlmostEqual(18230.6705, power[0], 4)
 
+  def test_get_air_conditioning_energy_rate_mismatched_lengths(self):
+    """ValueError when input vectors have different lengths."""
+    with self.assertRaisesRegex(
+        ValueError, 'All input vectors must be of the same length'
+    ):
+      energy_utils.get_air_conditioning_energy_rate(
+          air_flow_rates=[0.170, 0.180],  # 2 elements
+          outside_temps=[15 + 273.0],  # 1 element
```
s2t2: let's prefer to use the single kelvin temperature value 288 here, without performing addition (unless the addition is instructive somehow)

smart_control/utils/energy_utils_test.py
```diff
@@ -33,6 +33,35 @@ def test_get_humidity_ratio(self):
...
+    with self.assertRaisesRegex(
+        ValueError, 'Input arrays must have equal length'
+    ):
+      energy_utils.get_humidity_ratio(
+          temps=[293, 300],  # 2 elements
+          relative_humidities=[0.6],  # 1 element
+          pressures=[1.02],  # 1 element
+      )
+
+  def test_get_humidity_ratio_invalid_relative_humidity(self):
```
s2t2: it might be helpful to add tests for humidities and pressures where the values are zero (since that seems to be a notable edge case)

smart_control/utils/conversion_utils.py
```diff
@@ -64,22 +64,30 @@ def floor_plan_based_zone_identifier_to_id(identifier: str) -> str:
...
-  if m:
-    return int(m.group(1)), int(m.group(2))
-  raise ValueError('Could not convert zone_id to coordinates!')
+  # Expect exactly "zone_id_(<row>,<col>)" (optional spaces after comma)
+  m = re.match(r'^zone_id_\((\d+),\s*(\d+)\)$', zone_id)
+  if not m:
+    raise ValueError(
+        f"Invalid zone_id format: {zone_id!r}. Expected 'zone_id_(<row>,<col>)'"
+    )
+  return int(m.group(1)), int(m.group(2))
```
s2t2, line 73: nice, this looks like an improvement. thanks!

Gemini notes and explanation of regex change:

> 1. Regex Refinement
> The new regular expression is more precise.
>
> Standard Parentheses: The old version used [(] and [)] (character classes) to match parentheses. The new version uses \( and \), which is the standard way to escape literals in regex.
>
> Whitespace Flexibility: The addition of \s* after the comma is a great "quality of life" change. It allows the code to handle IDs like zone_id_(1,2) and zone_id_(1, 2) interchangeably.
>
> Strict Anchoring: Adding the $ at the end ensures the string matches exactly. The original would have accepted zone_id_(1, 2)random_extra_text.

smart_control/utils/conversion_utils_test.py
```diff
@@ -46,10 +46,34 @@ def test_normalize_hod(self):
     self.assertEqual(conversion_utils.normalize_hod(0), -1.0)
     self.assertEqual(conversion_utils.normalize_hod(23), 1.0)
 
+  def test_normalize_hod_invalid(self):
```
s2t2: Pro tip: here, let's implement an [absltest.parameterized test](https://byoshimi2-abseil-py.readthedocs.io/en/latest/absl.testing.parameterized.html), where we pass the params of 24 and -1. The main benefit is now these two separate tests can be run in parallel.

Perhaps something like this:

```py
from absl.testing import absltest
from absl.testing import parameterized

class ConversionUtilsTest(parameterized.TestCase):

  @parameterized.parameters(24, -1)
  def test_normalize_hod_invalid_raises_error(self, invalid_hod):
    with self.assertRaisesRegex(
        ValueError, r'Hour of day \(hod\) must be within the range \[0, 23\]'
    ):
      conversion_utils.normalize_hod(invalid_hod)

if __name__ == '__main__':
  absltest.main()
```

smart_control/utils/conversion_utils_test.py
```diff
@@ -46,10 +46,34 @@ def test_normalize_hod(self):
...
+    with self.assertRaisesRegex(
+        ValueError, r'Hour of day \(hod\) must be within the range \[0, 23\]'
+    ):
+      conversion_utils.normalize_hod(-1)
+
   def test_normalize_dow(self):
     self.assertEqual(conversion_utils.normalize_dow(0), -1.0)
     self.assertEqual(conversion_utils.normalize_dow(6), 1.0)
 
+  def test_normalize_dow_invalid(self):
```
s2t2: echo preference for parameterized test here. thank you!

smart_control/utils/regression_building_utils.py
```diff
@@ -384,7 +390,13 @@ def get_matching_indexes(
...
       output_indexes.append(ts_output)
 
-  assert len(output_indexes) == len(input_indexes)
+  if len(output_indexes) != len(input_indexes):
+    raise ValueError(
+        'Mismatch in matched input and output index lengths: '
+        f'input_indexes={len(input_indexes)}, '
+        f'output_indexes={len(output_indexes)}. '
+        'Matching logic failed to produce equal-length sequences.'
+    )
```
s2t2, line 399: Nice. It might be nice to add some tests for the two changes in this file.

smart_control/utils/controller_reader.py
```diff
@@ -149,9 +149,14 @@ def _select_shards(
...
-      ts = pd.Timestamp(re.findall(r'\d{4}\.\d{2}\.\d{2}\.\d{2}', filepath)[-1])
-      return ts
+      if not filepath:
+        raise ValueError('Filepath cannot be empty when reading timestamp.')
+      matches = re.findall(r'\d{4}\.\d{2}\.\d{2}\.\d{2}', filepath)
+      if not matches:
+        raise ValueError(
+            f'Could not extract timestamp from filepath: {filepath!r}'
+        )
+      return pd.Timestamp(matches[-1])
```
s2t2, line 159: I like the decision to separate this into multiple steps and check along the way. 

it might be nice to add some tests. however this is a private / protected method, so it could be ok to omit.

smart_control/simulator/vav.py
```diff
@@ -162,13 +165,20 @@ def compute_zone_supply_temp(
       supply_air_temp: Temperature in K of input air.
       input_water_temp: Temperature in K of input water.
     """
-    assert self.damper_setting > 0
-    assert self._max_air_flow_rate > 0
```
s2t2, line 166: let's restore these checks (as errors instead of assertions), to ensure positive values for both of these variables.

otherwise relying only on the `air_flow_rate` check below, it might be possible for these values to both be negative, and still pass. 

we can keep the `air_flow_rate` check for good measure.

smart_control/simulator/vav_test.py
```diff
@@ -131,6 +131,22 @@ def test_setters_raise_error(self):
     with self.assertRaises(ValueError):
       v.damper_setting = -0.1
 
+  def test_max_air_flow_rate_setter_raises_value_error(self):
+    """ValueError when max_air_flow_rate is set to 0 or negative."""
+    t = _get_default_thermostat()
+    b = _get_default_boiler()
+    v = vav.Vav(0.6, 0.4, t, b)
```
s2t2: let's prefer to use named parameters here please.

smart_control/simulator/vav_test.py
```diff
@@ -131,6 +131,22 @@ def test_setters_raise_error(self):
...
+
+    with self.assertRaisesRegex(
+        ValueError, 'Maximum air flow rate must be greater than 0'
+    ):
+      v.max_air_flow_rate = 0.0
+
+    with self.assertRaisesRegex(
+        ValueError, 'Maximum air flow rate must be greater than 0'
+    ):
+      v.max_air_flow_rate = -0.5
```
s2t2: let's use parameterized tests here (see related comment for more details / example)

s2t2 (changes requested):
@igopalakrishna thanks so much for this PR! 

Please see my comments requesting some minor revisions.

smart_control/reward/setpoint_energy_carbon_regret_test.py
igopalakrishna: Named parameters in setpoint_energy_carbon_regret_test.py -- done

smart_control/utils/energy_utils.py
igopalakrishna: if not (len(...) == len(...) == len(...)) style -- done

smart_control/utils/energy_utils.py
igopalakrishna: Numpy-based validation -- done (used np.where)

smart_control/simulator/randomized_arrival_departure_occupancy_test.py
igopalakrishna: Named parameters in randomized_arrival_departure_occupancy_test.py -- done

smart_control/reward/setpoint_energy_carbon_regret_test.py
igopalakrishna: FYI: Both calls now use named parameters (usd_per_kwh and kg_per_kwh) instead of positional arguments.

smart_control/simulator/randomized_arrival_departure_occupancy_test.py
igopalakrishna: FYI: All parameters are now named (earliest_expected_arrival_hour, latest_expected_arrival_hour, earliest_expected_departure_hour, latest_expected_departure_hour, step_size, random_state) instead of positional arguments.

smart_control/utils/energy_utils.py
igopalakrishna: FYI: if not (len(temps) == len(relative_humidities) == len(pressures)) matches the original implementation style.
The code also includes # pylint: disable=superfluous-parens to suppress the warning while keeping the requested style.

smart_control/utils/energy_utils.py
igopalakrishna: FYI: np.where() is a numpy-based method and fits here because it finds the first invalid index to include in the error message, which is more informative than np.all() or np.any()

smart_control/utils/energy_utils.py
igopalakrishna: FYI: The error message format (0,1] now matches the validation logic that excludes zero.

smart_control/utils/energy_utils_test.py
igopalakrishna: Done -- The test now uses direct Kelvin values (288 and 393) instead of performing addition, as requested.

smart_control/utils/energy_utils_test.py
igopalakrishna: FYI:The tests are parameterized, covering zero and other invalid values. Zero-value edge cases are covered for both relative humidity and pressure.

smart_control/utils/conversion_utils_test.py
igopalakrishna: The test is parameterized with 24 and -1, allowing parallel execution.

smart_control/utils/conversion_utils_test.py
igopalakrishna: Matched the pattern used for test_normalize_hod_invalid_raises_error. The test is now parameterized, allowing parallel execution of the test cases.

smart_control/simulator/vav.py
igopalakrishna: **Notes for Reference** 

### Verification:

- Both checks restored as ValueError (lines 170-178)
- Checks run before computing air_flow_rate (lines 180-183)
- The air_flow_rate check is kept (lines 186-191)
- Tests verify both cases (lines 255-263 in vav_test.py)

### Results:

- Prevents negative damper_setting and _max_air_flow_rate from passing
- Provides clear error messages
- The air_flow_rate check remains as a safeguard

The protobuf error in the test run is a pre-existing environment issue, not related to these changes.

smart_control/simulator/vav_test.py
igopalakrishna: The Vav constructor call now uses named parameters.

smart_control/simulator/vav_test.py
igopalakrishna: The test is now parameterized, allowing both test cases to run in parallel.

smart_control/utils/regression_building_utils.py
igopalakrishna: Thanks for the suggestion! These ValueError checks guard against internal logic errors that would be difficult to trigger in normal operation without mocking internal functions. Given the complexity of testing these edge cases and that these are defensive checks for internal consistency, I've left tests as optional for now. Happy to add them if you'd prefer!

smart_control/utils/controller_reader.py
igopalakrishna: Thanks for the positive feedback! Tests for this private method were intentionally omitted. The ValueError checks ensure robust error handling when timestamp extraction fails. Happy to add tests if you'd prefer.

smart_control/utils/conversion_utils.py
igopalakrishna: Thanks for the feedback! The stricter regex with whitespace flexibility should make the code more robust.

igopalakrishna:
@s2t2 Thanks for the detailed review! All requested changes have been addressed:

- Parameterized tests implemented
- Named parameters used throughout  
- Numpy methods used for validation
- Restored checks in vav.py
- Error messages updated
- All CI checks passing

Ready for re-review!

smart_control/utils/energy_utils.py
```diff
@@ -63,7 +63,33 @@ def get_humidity_ratio(
 
   Returns: water mass to air mass ratio in kg Water / kg Air
   """
-  assert len(temps) == len(relative_humidities) == len(pressures)
+  if not (len(temps) == len(relative_humidities) == len(pressures)):  # pylint: disable=superfluous-parens
```
s2t2, line 66: LOL sorry I recommended superfluous parens 👻 

(no change required, can leave like this)

smart_control/utils/energy_utils.py
```diff
@@ -63,7 +63,33 @@ def get_humidity_ratio(
...
+
+  # Sanity-check each RH and pressure using numpy
+  relative_humidities_array = np.array(relative_humidities)
+  pressures_array = np.array(pressures)
+
+  invalid_rh_indices = np.where(
+      (relative_humidities_array <= 0.0) | (relative_humidities_array > 1.0)
+  )[0]
+  if len(invalid_rh_indices) > 0:
+    i = invalid_rh_indices[0]
```
s2t2: if I'm reading this correctly, if there are any invalid values, we will show the first one in the error message? 

for readability, let's prefer to simplify, perhaps using something like this:

```
if np.any((relative_humidities_array <= 0.0) | (relative_humidities_array > 1.0)):
    raise ValueError("Relative humidities must be in the range (0, 1].")
```

we can omit inclusion of the invalid values themselves in the error message, and let the developer debug to find out more details.

smart_control/utils/energy_utils_test.py
```diff
@@ -33,16 +33,67 @@ def test_get_humidity_ratio(self):
...
+        ValueError, 'Input arrays must have equal length'
+    ):
+      energy_utils.get_humidity_ratio(
+          temps=[293, 300],  # 2 elements
+          relative_humidities=[0.6],  # 1 element
+          pressures=[1.02],  # 1 element
+      )
+
+  @parameterized.parameters(
+      (1.5, r'Relative humidities must be in \(0,1\]'),
```
s2t2: since all three cases use the same error message, let's remove duplication by referencing the error message once, inside the test, like:

```
with self.assertRaisesRegex(ValueError, r'Relative humidities must be in \(0,1\]'):
```

s2t2: this applies for the pressures test below as well

smart_control/utils/energy_utils.py
s2t2: echo for the pressures check below as well.

```
if np.any(pressures_array <= 0.0):
    raise ValueError("Pressures must be greater than 0 (bar).")
```

s2t2 (changes requested):
@igopalakrishna thanks for the quick turn around!

we are getting close. I just have a few more final comments.

smart_control/utils/energy_utils.py
igopalakrishna: Done! Simplified the validation using np.any() and removed specific invalid values from error messages. Applied to both relative_humidities and pressures checks. All tests passing.

smart_control/utils/energy_utils_test.py
igopalakrishna: Done! Error messages are now referenced once inside each test, not duplicated in parameters. Applied to both relative_humidities and pressures tests.

igopalakrishna:
Thank you @s2t2! All final comments addressed:

- Simplified validation using np.any() and removed invalid values from error messages
- Applied to both relative_humidities and pressures checks
- Error messages in tests are now inside the test (not duplicated)
- All tests passing

Ready for final review!

s2t2 (approved):
LGTM 🚀

PR #114, opened 2025-07-27
refactor: Modernize super() calls in sac_networks 

### LINKED ISSUE #35: Check Inheritance of `_TanhNormalProjectionNetworkWrapper`
We see `_TanhNormalProjectionNetworkWrapper` is calling super on itself? Is this intended? If so, can we add some context to the docstrings?

https://github.com/google/sbsim/blob/ca40d7b3edd758ab9c0025ee2b0f23993b5d5594/smart_control/reinforcement_learning/agents/networks/sac_networks.py#L112

https://github.com/google/sbsim/blob/ca40d7b3edd758ab9c0025ee2b0f23993b5d5594/smart_control/reinforcement_learning/agents/networks/sac_networks.py#L119

Or are we meaning to call super on the parent class `tanh_normal_projection_network.TanhNormalProjectionNetwork` from tf-agents? If so, can we update this code, and potentially also remove the class name from the `super()` call?

@gabriel-trigo

igopalakrishna:
### Summary

This PR addresses issue **#35** by refactoring the `_TanhNormalProjectionNetworkWrapper` class to use the modern, argument-less `super()` syntax. This change improves code readability and aligns the codebase with current Python 3 best practices without altering functionality.

### Technical Explanation

The previous implementation in `sac_networks.py` used the `super(_TanhNormalProjectionNetworkWrapper, self)` syntax. While functionally correct, this is an explicit and verbose style inherited from Python 2. It can be confusing as it appears to call `super()` on the class itself.

The goal of `super()` is to call the method from the parent class in the Method Resolution Order (MRO), which in this case is `tanh_normal_projection_network.TanhNormalProjectionNetwork`. The modern `super()` call automatically handles this, making the code cleaner, more maintainable, and less prone to copy-paste errors if the class were ever to be renamed.

This refactoring replaces the outdated calls with `super()`, which is the idiomatic standard in Python 3.

### Changes Made

- Updated the `__init__` method in `_TanhNormalProjectionNetworkWrapper` to use `super().__init__(...)`.
- Updated the `call` method in `_TanhNormalProjectionNetworkWrapper` to use `super().call(...)`.

### Validation 

To ensure this refactoring did not introduce any regressions, the full project test suite was executed using `pytest`.

- **Test Command:** `pytest`
- **Result:** **671 passed, 38 skipped, 0 failed**

The successful test run validates that this change is safe and does not impact existing behavior.

Fixes #35

- [x] Tests pass
- [x] No documentation changes were necessary for this change.

### DIFF
--- a/smart_control/reinforcement_learning/agents/networks/sac_networks.py
+++ b/smart_control/reinforcement_learning/agents/networks/sac_networks.py
@@ -109,14 +109,14 @@ class _TanhNormalProjectionNetworkWrapper(
   """Wrapper to pass predefined `outer_rank` to underlying projection net."""
 
   def __init__(self, sample_spec, predefined_outer_rank=1):
-    super(_TanhNormalProjectionNetworkWrapper, self).__init__(sample_spec)
+    super().__init__(sample_spec)
     self.predefined_outer_rank = predefined_outer_rank
 
   def call(self, inputs, **kwargs):
     kwargs['outer_rank'] = self.predefined_outer_rank
     if 'step_type' in kwargs:
       del kwargs['step_type']
-    return super(_TanhNormalProjectionNetworkWrapper, self).call(
+    return super().call(
         inputs, **kwargs
     )
 


ajkhabbazi:
@s2t2 I reviewed the diff, this is a no-behavior-change refactor that replaces `super(_TanhNormalProjectionNetworkWrapper, self)` with Python 3 idiomatic `super()`, matching the intent of issue #35. All CI checks are green and there are no merge conflicts. LGTM; should be safe to merge.

s2t2:
@igopalakrishna thank you for the fix, and @ajkhabbazi thank you for the review!

Merging... 🚀 🚀 ✨

PR #98, opened 2025-06-11
Reinforcement Learning Module, Part 2

gabriel-trigo:
Updates the reinforcement learning module to include additional functionality.

Details:
- Adds "eval.py" script to evaluate policies.
- Adds "generate_gin_config_files.py" script to generate variations of gin environment config files.
- Adds visualization module with features to save plots of evaluation results.
- Adds implementation of td3 and ddpg agents.
- Removes ambiguity between the Environment's `step_interval` and the BaseBuilding's `time_step_sec` property. Fixes #20.
- Refactors filepath constants and temperature conversion functions. Closes #25.

PR #91, opened 2025-06-05
Fix Spelling Issues

codewithbear:
Fix for https://github.com/google/sbsim/issues/53

- AI (openai/gpt-4.1):
```
Fix spelling mistakes in documentation and python comments. Do not fix correct regional variations such as British English.
** Important do not fix spelling in variable names **
```
 (Single file: Yes)

Generated by Morph [codemorph.dev](https://codemorph.dev)

PR #123, opened 2025-08-29
Enhanced Occupancy Model 

amcberkes:
Added enhanced occupancy model with minute-level control and different occupant types to model weekend behaviour, included daily parameter sampling for realistic behaviour that changes from day to day. Added test file for the new occupancy model. Fixed bug in `randomized_arrival_departure_occupancy.py`.

## Summary
This PR introduces an enhanced occupancy model that provides more realistic building occupancy simulation with the following improvements:
### New Features
- Enhanced occupancy model (`smart_control/simulator/enhanced_occupancy.py`):
- Minute-level temporal control 
- Different occupant types to accurately model weekend behaviour patterns
- Daily parameter sampling for realistic day-to-day behavioural variation
- Tests (`smart_control/simulator/enhanced_occupancy_test.py`)

### Bug Fixes
- Fixed bug in `randomized_arrival_departure_occupancy.py`: average_zone_occupancy() returned only an instantaneous count instead of computing the true time-averaged occupancy over the specified interval

### Other Changes  
- Updated GitHub issue and pull request templates

## Testing
All pre-commit hooks pass, as well as the tests in `enhanced_occupancy_test.py` and `randomized_arrival_departure_occupancy_test.py`

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,4 +1,4 @@
-Fixes #<issue_number_goes_here>
+Fixes #\<issue_number_goes_here>
 
 > It's a good idea to open an issue first for discussion.
 
--- a/smart_control/simulator/enhanced_occupancy.py
+++ b/smart_control/simulator/enhanced_occupancy.py
@@ -1,0 +1,416 @@
+"""A enhanced stochastic occupancy model for building simulation
+with minute-level control and different worker types.This model is based on the
+LIGHTSWITCHOccupancy model from stochastic_occupancy.py to include minute-level
+control instead of hour-level control. It has the same arrival/departure/lunch
+logic but it provides more fine-grained control and realistic occupant
+behaviour, such as optional weekend work and daily changing work hours and lunch
+break times (instead of a constant set of parameters for each occupant leading
+to a static occupancy profile that repreats itself every weekday of the year).
+The model samples new work and lunch time parameters for each occupant every
+day, caches them for the day and clears the cache on the next day to ensure
+consistency. It is possible to model low occupancy levels on the weekends by
+using different worker types.
+"""
+
+import datetime
+import enum
+from typing import Dict, Union
+
+import gin
+import numpy as np
+import pandas as pd
+
+from smart_control.models.base_occupancy import BaseOccupancy
+from smart_control.utils import conversion_utils
+
+debug_print = False  # Set to False to disable debugging
+
+
+class OccupancyStateEnum(enum.Enum):
+  AWAY = 1
+  WORK = 2
+
+
+class WorkerType(enum.Enum):
+  WEEKDAY_ONLY = 1
+  WEEKEND_REGULAR = 2  # worker types that works on weekends and weekdays
+  WEEKEND_OCCASIONAL = (
+      3  # regular weekday workers who need to work on weekends occasionally
+  )
+
+
+class EnhancedZoneOccupant:
+  """EnhancedZone Occupant with minute-level contril and day-specific
+  parameters."""
+
+  def __init__(
+      self,
+      earliest_expected_arrival_min: int,
+      latest_expected_arrival_min: int,
+      earliest_expected_departure_min: int,
+      latest_expected_departure_min: int,
+      lunch_start_min: int,
+      lunch_end_min: int,
+      step_size: pd.Timedelta,
+      random_state: np.random.RandomState,
+      time_zone: Union[datetime.tzinfo, str] = "UTC",
+      worker_type: WorkerType = WorkerType.WEEKDAY_ONLY,
+      weekend_work_prob: float = 0.10,
+      occupant_id: int = 0,
+  ):
+    assert (
+        earliest_expected_arrival_min
+        < latest_expected_arrival_min
+        < earliest_expected_departure_min
+        < latest_expected_departure_min
+    )
+    assert lunch_start_min < lunch_end_min
+
+    self._earliest_expected_arrival_min = earliest_expected_arrival_min
+    self._latest_expected_arrival_min = latest_expected_arrival_min
+    self._earliest_expected_departure_min = earliest_expected_departure_min
+    self._latest_expected_departure_min = latest_expected_departure_min
+    self._lunch_start_min = lunch_start_min
+    self._lunch_end_min = lunch_end_min
+    self._step_size = step_size
+    self._random_state = random_state
+    self._time_zone = time_zone
+    self.state = OccupancyStateEnum.AWAY
+    self.daily_cache = {}
+    self.worker_type = worker_type
+    self.weekend_work_prob = weekend_work_prob
+    self.occupant_id = occupant_id
+    self.id = occupant_id
+    self.daily_work_cache = {}
+
+  def _generate_cpf(self, start, end, random_state=None):
+    if random_state is None:
+      random_state = self._random_state
+    values = np.arange(start, end + 1)
+    probabilities = random_state.rand(len(values))
+    cumulative_probabilities = np.cumsum(probabilities / probabilities.sum())
+    return values, cumulative_probabilities
+
+  def _sample_event_time(self, start, end, random_state=None):
+    if random_state is None:
+      random_state = self._random_state
+    values, cumulative_probabilities = self._generate_cpf(
+        start, end, random_state
+    )
+    random_value = random_state.rand()
+    index = np.searchsorted(cumulative_probabilities, random_value)
+    if debug_print:
+      print(
+          f"Sampled event time: start={start}, end={end}, value={values[index]}"
+      )
+    return values[index]
+
+  def _sample_lunch_duration(self, random_state=None):
+    if random_state is None:
+      random_state = self._random_state
+    values, cumulative_probabilities = self._generate_cpf(30, 90, random_state)
+    random_value = random_state.rand()
+    index = np.searchsorted(cumulative_probabilities, random_value)
+    if debug_print:
+      print(f"Sampled lunch duration: {values[index]} minutes")
+    return values[index]
+
+  def _to_local_time(self, timestamp: pd.Timestamp) -> pd.Timestamp:
+    if timestamp.tz is None:
+      if self._time_zone is None:
+        return timestamp
+      return timestamp.tz_localize(self._time_zone)
+    return timestamp.tz_convert(self._time_zone)
+
+  def _get_daily_params(
+      self, timestamp: pd.Timestamp, local_timestamp: pd.Timestamp = None
+  ) -> Dict[str, int]:
+    if local_timestamp is None:
+      local_timestamp = self._to_local_time(timestamp)
+
+    date_key = local_timestamp.date()
+
+    if self.daily_cache and list(self.daily_cache.keys())[0] != date_key:
+      self.daily_cache.clear()
+
+    if date_key in self.daily_cache:
+      return self.daily_cache[date_key]
+
+    day_seed = hash(str(date_key) + str(self.occupant_id) + "daily_params") % (
+        2**32
+    )
+    day_random_state = np.random.RandomState(day_seed)
+
+    arrival = self._sample_event_time(
+        self._earliest_expected_arrival_min,
+        self._latest_expected_arrival_min,
+        day_random_state,
+    )
+    departure = self._sample_event_time(
+        self._earliest_expected_departure_min,
+        self._latest_expected_departure_min,
+        day_random_state,
+    )
+    lunch_start = self._sample_event_time(
+        self._lunch_start_min, self._lunch_end_min, day_random_state
+    )
+    lunch_duration = self._sample_lunch_duration(day_random_state)
+
+    self.daily_cache[date_key] = {
+        "arrival_time": arrival,
+        "departure_time": departure,
+        "lunch_start_time": lunch_start,
+        "lunch_duration": lunch_duration,
+    }
+    return self.daily_cache[date_key]
+
+  def _minutes_since_midnight(self, local_timestamp: pd.Timestamp) -> int:
+    return local_timestamp.hour * 60 + local_timestamp.minute
+
+  def _should_work_today(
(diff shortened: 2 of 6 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,6 +1,6 @@
-Fixes #<issue_number_goes_here>
+Fixes #\<issue_number_goes_here>
```
s2t2, line 1: I think these automatic formatting updates are fine. We have some similar updates coming to these files in #98 . I may take a look at updating the tooling in the future to prevent auto-formatting of markdown files in the ".github" directory.

smart_control/simulator/enhanced_occupancy.py
```diff
@@ -0,0 +1,416 @@
+"""A enhanced stochastic occupancy model for building simulation
+with minute-level control and different worker types.This model is based on the
```
s2t2: let's add a space between "types." and "This"

smart_control/simulator/enhanced_occupancy.py
```diff
@@ -0,0 +1,416 @@
...
+LIGHTSWITCHOccupancy model from stochastic_occupancy.py to include minute-level
+control instead of hour-level control. It has the same arrival/departure/lunch
+logic but it provides more fine-grained control and realistic occupant
+behaviour, such as optional weekend work and daily changing work hours and lunch
+break times (instead of a constant set of parameters for each occupant leading
+to a static occupancy profile that repreats itself every weekday of the year).
+The model samples new work and lunch time parameters for each occupant every
+day, caches them for the day and clears the cache on the next day to ensure
+consistency. It is possible to model low occupancy levels on the weekends by
+using different worker types.
```
s2t2, line 12: Appreciate the detailed explanation in the module docstring. 🙌

smart_control/simulator/enhanced_occupancy.py
```diff
@@ -0,0 +1,416 @@
...
+from typing import Dict, Union
+
+import gin
+import numpy as np
+import pandas as pd
+
+from smart_control.models.base_occupancy import BaseOccupancy
+from smart_control.utils import conversion_utils
+
+debug_print = False  # Set to False to disable debugging
```
s2t2: Let's prefer to replace all print statements with INFO level logging. Then whoever is running the file can choose their logging level. That will help get rid of this manual print toggle as well. 

Example:

```
from absl import logging

logging.info("blah blah blah")
```

smart_control/simulator/enhanced_occupancy.py
```diff
@@ -0,0 +1,416 @@
...
+class WorkerType(enum.Enum):
+  WEEKDAY_ONLY = 1
+  WEEKEND_REGULAR = 2  # worker types that works on weekends and weekdays
+  WEEKEND_OCCASIONAL = (
+      3  # regular weekday workers who need to work on weekends occasionally
+  )
+
+
+class EnhancedZoneOccupant:
+  """EnhancedZone Occupant with minute-level contril and day-specific
```
s2t2: I think we want to update spelling from "contril" to "control". 

We also might want to add or remove a space to make it either `EnhancedZoneOccupant` or `Enhanced Zone Occupant` for consistency (unless there is a concept of an `EnhancedZone`).

Also when ending a multi-line docstring, let's consider putting the final triple quote on its own line below.

smart_control/simulator/enhanced_occupancy.py
```diff
@@ -0,0 +1,416 @@
...
+      lunch_start_min: int,
+      lunch_end_min: int,
+      step_size: pd.Timedelta,
+      random_state: np.random.RandomState,
+      time_zone: Union[datetime.tzinfo, str] = "UTC",
+      worker_type: WorkerType = WorkerType.WEEKDAY_ONLY,
+      weekend_work_prob: float = 0.10,
+      occupant_id: int = 0,
+  ):
+    assert (
```
s2t2: Let's prefer to raise a `ValueError` instead of using assertions in non-test code. This is something we are working on cleaning from the rest of the codebase in #28 .

smart_control/simulator/enhanced_occupancy.py
```diff
@@ -0,0 +1,416 @@
...
+    if debug_print:
+      print(
+          f"Sampled event time: start={start}, end={end}, value={values[index]}"
+      )
+    return values[index]
+
+  def _sample_lunch_duration(self, random_state=None):
+    if random_state is None:
+      random_state = self._random_state
+    values, cumulative_probabilities = self._generate_cpf(30, 90, random_state)
```
s2t2: Let's consider setting 30 and 90 as default parameters for this function, so they can be customized if desired.

smart_control/simulator/enhanced_occupancy.py
```diff
@@ -0,0 +1,416 @@
...
+    random_value = random_state.rand()
+    index = np.searchsorted(cumulative_probabilities, random_value)
+    if debug_print:
+      print(f"Sampled lunch duration: {values[index]} minutes")
+    return values[index]
+
+  def _to_local_time(self, timestamp: pd.Timestamp) -> pd.Timestamp:
+    if timestamp.tz is None:
+      if self._time_zone is None:
```
s2t2: Would it make sense to enforce the presence of a time zone value for this class? Perhaps we can add a validation to the constructor to ensure a (valid) time zone is passed in?

Then we can remove this check to see whether the time zone is set, and this function will always return a timezone-aware timestamp. Otherwise there would be some conditions where this function returns timezone-aware timestamps and other conditions where it returns timezone-naive timestamps, and that might complicate things.

Maybe I am missing something. Let me know what you think.

smart_control/simulator/enhanced_occupancy.py
```diff
@@ -0,0 +1,416 @@
...
+    date_key = local_timestamp.date()
+
+    if self.daily_cache and list(self.daily_cache.keys())[0] != date_key:
+      self.daily_cache.clear()
+
+    if date_key in self.daily_cache:
+      return self.daily_cache[date_key]
+
+    day_seed = hash(str(date_key) + str(self.occupant_id) + "daily_params") % (
+        2**32
```
s2t2: it might be nice to add a comment to explain the practical effects of this `2**32` modulo operation, if possible

smart_control/simulator/enhanced_occupancy.py
s2t2: Not the most important (perhaps a nitpick), but in terms of naming things, I wonder if there are some alternative class names we can consider. For example something like `MinuteLevelZoneOccupant` or something like that captures the main functionality differences between this and the previous / existing version. Otherwise totally fine to leave the name as is.

smart_control/simulator/enhanced_occupancy.py
s2t2: On a related note, in a few different functions (for example in `_get_daily_params` below) we see two different parameters for the timestamp concept, and this code to convert the timezone-naive one to be timezone-aware:

```py
def _get_daily_params(self, timestamp: pd.Timestamp, local_timestamp: pd.Timestamp = None):
  if local_timestamp is None:
    local_timestamp = self._to_local_time(timestamp)
```

However is there a way to simplify?

One idea is that we can just pass a single timestamp parameter to these functions, and then unconditionally force it into the class' timezone. Perhaps using an approach like this:

```py
# simplifying the signature (uses one param instead of two):
def _get_daily_params(self, timestamp: pd.Timestamp): 
    timestamp = self._to_local_time(timestamp) # force timezone aware
    # ... use the timestamp
```

Let me know what you think, or if I'm missing something.

smart_control/simulator/enhanced_occupancy.py
s2t2: On a related note, we see the same `2**32` used in various hashing implementations throughout the file. Would it be helpful to abstract this value away into its own constant at the top of the file?

smart_control/simulator/enhanced_occupancy.py
```diff
@@ -0,0 +1,416 @@
...
+        if state == OccupancyStateEnum.WORK:
+          num_occupants += 1
+
+      total_occupancy += num_occupants
+      steps += 1
+      current_time += self._step_size
+
+    return total_occupancy / steps if steps > 0 else 0.0
+
+  def get_worker_distribution(self, zone_id: str) -> Dict[str, int]:
```
s2t2, line 446: Nice, this function will be helpful. Let's consider adding a docstring (Google internal codebase will require a docstring for all public methods).

smart_control/simulator/enhanced_occupancy.py
```diff
@@ -0,0 +1,416 @@
...
+
+class WorkerType(enum.Enum):
+  WEEKDAY_ONLY = 1
+  WEEKEND_REGULAR = 2  # worker types that works on weekends and weekdays
+  WEEKEND_OCCASIONAL = (
+      3  # regular weekday workers who need to work on weekends occasionally
+  )
+
+
+class EnhancedZoneOccupant:
```
s2t2: Does it make sense to inherit this from `ZoneOccupant` in "randomized_arrival_departure_occupancy.py"?

I see this class is using minutes while the other is using hours. So maybe that is difference enough to not inherit. Maybe we can add a brief note to the docstring to explain our choice.

smart_control/simulator/enhanced_occupancy.py
```diff
@@ -0,0 +1,416 @@
...
+    if debug_print:
+      print(f"Occupancy state: {self.state}")
+
+    return self.state
+
+
+@gin.configurable
+class EnhancedOccupancy(BaseOccupancy):
+  """Enhanced occupancy model with minute-level control and different
+  worker types."""
```
s2t2: echo earlier comments about docstring formatting. Let's move the ending `"""` to its own line below.

smart_control/simulator/randomized_arrival_departure_occupancy.py
```diff
@@ -197,8 +197,15 @@ def average_zone_occupancy(
...
+    steps = 0
+    while current_time < end_time:
+      num_occupants = 0.0
+      for occupant in self._zone_occupants[zone_id]:
+        if occupant.peek(current_time) == OccupancyStateEnum.WORK:
+          num_occupants += 1.0
+      total_occupants += num_occupants
+      steps += 1
+      current_time += self._step_size
+    return total_occupants / steps if steps > 0 else 0.0
```
s2t2, line 211: Thanks for this update! Is there a test we can add for the randomized arrival departure occupancy class' `average_zone_occupancy` method to capture the change? For example one that would fail under the previous behavior but passes with the updated behavior?

smart_control/simulator/enhanced_occupancy_test.py
```diff
@@ -0,0 +1,259 @@
...
+    )
+    day1_early_morning = pd.Timestamp('2021-09-01 06:00')
+    day1_work_morning = pd.Timestamp('2021-09-01 10:00')
+    day1_afternoon = pd.Timestamp('2021-09-01 15:00')
+    day1_evening = pd.Timestamp('2021-09-01 20:00')
+    weekend = pd.Timestamp('2021-09-05 08:00')
+
+    self.assertEqual(occupant.peek(day1_early_morning), OccupancyStateEnum.AWAY)
+    self.assertEqual(occupant.peek(day1_work_morning), OccupancyStateEnum.WORK)
+    self.assertEqual(occupant.peek(day1_afternoon), OccupancyStateEnum.WORK)
```
s2t2: looks like there might be a test failure around here. please check.

https://github.com/google/sbsim/actions/runs/17325401231/job/49452306954?pr=123#step:11:208

I see one of the builds [passed](https://github.com/google/sbsim/actions/runs/17325401231/job/49452306955?pr=123) though. Could be an ephemeral failure?

s2t2 (changes requested):
Hi @amcberkes thank you for this PR! 🙌 Looks like it will be a nice addition that will provide more complex and realistic occupancy modeling.

I left a lot of review comments. 

The most important ones are: 

  + a suggestion to enforce / validate the presence of timezone for the `EnhancedZoneOccupant` class ([comment](https://github.com/google/sbsim/pull/123/files#r2317446038)), which may also allow us to simplify some of the method signatures accordingly to accept a single timestamp parameter instead of two
  + replacing printing with logging ([comment](https://github.com/google/sbsim/pull/123/files#r2317433902))
  + changing assertions in non-test code to error raising instead ([comment](https://github.com/google/sbsim/pull/123/files#r2317436955))
  + fixing failing test(s) ([comment](https://github.com/google/sbsim/pull/123/files#r2317515372))

Please feel free to reply to any of the comments if you would like to clarify or discuss further.

Otherwise let me know when you have finished implementing some updates and are ready for another review.

.github/PULL_REQUEST_TEMPLATE.md
amcberkes: Thanks for clarifying! I wasn’t sure whether to commit those auto-formatting changes, but I’ll leave them in since you said it's fine.

smart_control/simulator/enhanced_occupancy.py
amcberkes: done

smart_control/simulator/enhanced_occupancy.py
amcberkes: thank you :)

amcberkes:
Hi @s2t2 , thanks for the detailed review! I have pushed an updated version with your suggestions. Lmk what you think!

s2t2 (approved):
@amcberkes thank you for implementing the requested changes. This PR is looking good to me. And ready for merge. 
 
FYI - there is a pipeline of incoming PRs (both internal and open source), and in hopes of minimizing merge conflicts for various contributors, I had been thinking of merging the PRs in the following order: #94 first, then #91, and then this one. However I am alternatively considering merging the PRs that are ready first. So stay tuned, but we will hope to get this merged as soon as possible.

amcberkes:
@s2t2 thanks for the update! Sounds good.

PR #124, opened 2025-09-24
Rad default table update

ecosang:
@s2t2 
I've updated the table. Please review if the format is correct for mkdocs.


s2t2:
@ecosang thanks! looks like there are some old changes in this PR - can you please rebase? or check out the Copybara push branch, pull the most recent upstream changes, and create a new branch and update the table there?

PR #125, opened 2025-09-30
Fix Radiation Properties Table

ecosang:
Updates the documentation about radiation properties, fixing the table of benchmark values for common materials.

### DIFF
--- a/smart_control/simulator/building.py
+++ b/smart_control/simulator/building.py
@@ -74,39 +74,27 @@ class RadiationProperties:
   Each of the property values should be between 0 and 1 (inclusive). Example
   values for various common materials are displayed in the tables below.
 
-  Natural Surfaces:
+  Table 4.5 Long-wave and solar emissivity for building surfaces:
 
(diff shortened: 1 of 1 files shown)


ecosang:
@s2t2 My apologies — I think I forgot to pull from my origin before making changes. When I created the PR, everything got bundled together, which seemed really weird at first. I assumed Git would somehow track everything properly, but it turns out it was my mistake.

It's strange because I thought I had pulled after syncing, but clearly, something went wrong. Thanks for catching it!

This is one of my first PRs to an open source project, so I'm still getting familiar with the process. I appreciate your patience.

s2t2:
@ecosang all good. I previewed this table with `make docs` locally and updated the formatting slightly. It is looking good and ready for merge.

We are seeing some weird unrelated build errors on Mac due to underlying system dependencies. I will try to re-run the Mac build a few times to ideally get it to pass.

---

Screenshot From 2025-09-30 17-21-15

PR #131, opened 2025-11-11
Add interior mass nodes

ecosang:
image
After discussion with John and Donghun, I have added interior mass into the current sbsim's 2D formulation.
As shown in the above figure, the main idea is to add interior mass nodes (adiabatic to the surroundings except for the each vertical air node). By doing so, we can include absorbed heat into the indoor space (as air node's thermal capacitance is too small).

After adding this feature, you can now see a greater temperature gradient due to the interior thermal mass.
image

I've implemented it for the iterative approach and the tensorized approach. Test cases added, and the formulations are included in the docstrings. 
Finally, the calculation is validated via tf_simulator_test.py by comparing with the iterative approach. One thing I found is that when the convergence threshold is too small (1e-4), the convergence takes lots of time as the interior mass's temperature change is very slow.

Also, some minor fixes for interior radiative heat exchange, such as naming convention (capital letters to small letters).
Hardcoded expected results are converted to .csv file in test cases.


ecosang:
@s2t2  I can't assign reviewers. Please see my above responses.

s2t2:
@ecosang awesome! I will aim to review this week.

smart_control/simulator/conftest.py
```diff
@@ -53,6 +54,9 @@ def create_building_with_radiative_properties(
   building_exterior_properties = building.MaterialProperties(
       conductivity=0.05, heat_capacity=500.0, density=3000.0
   )
+  interior_mass_properties = building.MaterialProperties(
+      conductivity=0.5, heat_capacity=1000.0, density=2000.0
```
s2t2, line 58: for the `MaterialProperties` class, it would be great if we could add a value lookup table in the docs (like we did for the `RadiationProperties`). 

I know the `MaterialProperties` is not a new class, but I was wondering if you have any insight into what some default values would be for common materials.

smart_control/simulator/building.py
```diff
@@ -846,6 +856,9 @@ def __init__(
     self._convection_simulator = convection_simulator
     self._reset_temp_values = reset_temp_values
     self.include_radiative_heat_transfer = include_radiative_heat_transfer
+    self.include_interior_mass = include_interior_mass
+    # self._interior_mass_properties = interior_mass_properties
+    # self._inside_air_properties = inside_air_properties
```
s2t2: please remove these commented lines before we merge

smart_control/simulator/building.py
```diff
@@ -801,6 +804,9 @@ def __init__(
       building_exterior_radiative_properties: RadiationProperties | None = None,
       include_radiative_heat_transfer: bool = False,
       view_factor_method: str = "ScriptF",
+      interior_mass_properties: MaterialProperties | None = None,
+      interior_mass_radiative_properties: RadiationProperties | None = None,
```
s2t2: I believe we may need to also add `interior_mass_radiative_properties` to the class docstring

smart_control/simulator/building.py
```diff
@@ -881,6 +894,11 @@ def __init__(
     if zone_map is None and zone_map_filepath is not None:
       zone_map = building_utils.read_floor_plan_from_filepath(zone_map_filepath)
       self._zone_map = zone_map
+    if include_interior_mass and interior_mass_properties is None:
```
s2t2, line 930: I was initially wondering if we also need to check for `interior_mass_radiative_properties` here. 

However I saw below we are using default values of `DefaultInsideWallRadiationProperties`, so I think this is OK.

smart_control/simulator/building.py
```diff
@@ -1020,25 +1051,101 @@ def _assign_radiative_heat_transfer_properties(
...
+          self.interior_wall_vf, a_tilde_inv
       )
 
     else:
       self.view_factor_method = None
       self.indexed_floor_plan = None
       self.interior_wall_mask = None
       self.interior_wall_index = None
-      self.interior_wall_VF = None
+      self.interior_wall_vf = None
```
s2t2, line 1111: I like the stylistic changes of lowercasing these variable names. It moves us further towards using descriptive variable names for these mathematical notation variables.

> Prefer PEP8-compliant descriptive names for public APIs...

https://google.github.io/styleguide/pyguide.html#3165-mathematical-notation

smart_control/simulator/building.py
```diff
@@ -801,6 +804,9 @@ def __init__(
       building_exterior_radiative_properties: RadiationProperties | None = None,
       include_radiative_heat_transfer: bool = False,
       view_factor_method: str = "ScriptF",
+      interior_mass_properties: MaterialProperties | None = None,
```
s2t2: It might be nice to move the `interior_mass_properties`  under the existing material properties, and `interior_mass_radiative_properties` down below the existing radiation properties. This would help group like concepts together (keeping all material properties together, and keeping all radiation properties together).

smart_control/simulator/building.py
```diff
@@ -1056,6 +1163,18 @@ def conductivity(self) -> np.ndarray:
...
+    return self._interior_mass_conductivity
+
+  @property
+  def interior_mass_heat_capacity(self) -> np.ndarray:
+    return self._interior_mass_heat_capacity
+
+  @property
+  def interior_mass_density(self) -> np.ndarray:
+    return self._interior_mass_density
+
```
s2t2, line 1210: About these three new interior mass properties... I was initially wondering how they differ from the related existing properties  `conductivity`,  `heat_capacity`, and `density`, respectively. We see the existing properties getting set using the `_assign_interior_and_exterior_values` method which takes inputs relating to the wall and inside air control volumes. 

Perhaps we can add something like this to the documentation about interior mass, perhaps in the class Docstring, to help explain how it differs conceptually from control volumes:

>  Interior Mass refers to the collection of objects and materials within the conditioned spaces (i.e., rooms) of the building that are not part of the building structure itself (like walls, floors, ceilings). These objects possess thermal mass, meaning they can absorb, store, and release heat.

s2t2 (commented):
Hi @ecosang I have conducted a preliminary review with some comments that I will release now, however I still need to spend some more time to complete my review, which I aim to do next Tuesday.

smart_control/simulator/tf_simulator.py
```diff
@@ -563,19 +568,196 @@ def _get_tensor_exterior_mask(
...
   ) -> tuple[np.ndarray, float]:
-    """Iterates across all CVs and updates the temperature estimate.
+    r"""Iterates across all CVs and updates the temperature estimate.
 
     Corner and edge CVs are exposed to thermal exchange with the ambient air
-    through convection.
-
-    This method implements Equation 22, derived in
-    go/smart-buildings-simulator-design.
+    through convection. This tensorized implementation overrides the parent
```
ecosang, line 709: Do you know if this docstring shows up in documentation? @s2t2

smart_control/simulator/building.py
ecosang: I will review this. I don't want to make a change for the original method `_assign_interior_and_exterior_values` and also I felt it is a bit hard to track. 
One thing is as a default, we only need to use the original cond+conv heat transfer. Then, we can choose either to enable radiative_heat_transfer+interior mass.  When there are not much solar radiation, interior_mass can be skipped (as it increases the # iteration). I will take a look at, but at this point, I think it is better to split it because interior_mass is still optional.

smart_control/simulator/conftest.py
ecosang: The original material properties are designed by Google team, and I made radiative property with you.. so I will find out default values and then make them consistent.

smart_control/simulator/building.py
s2t2: OK that sounds reasonable. Thanks for sharing your thought process.

smart_control/simulator/tf_simulator.py
s2t2: Yes I believe it is displayed here:
https://google.github.io/sbsim/api/simulator/simulator/#smart_control.simulator.tf_simulator.TFSimulator.update_temperature_estimates

NOTE: if you would like us to move or reorganize any part of the docs, we can definitely do that.  If a different organizational structure makes sense let's definitely update the locations of various classes in the docs, as desired.

smart_control/simulator/building_radiation_utils.py
```diff
@@ -46,7 +47,7 @@ def calculate_A_tilde_inv(epsilon: np.ndarray, F: np.ndarray) -> np.ndarray:
   return np.linalg.inv(A)
 
 
-def calculate_IFAinv(F: np.ndarray, A_inv: np.ndarray) -> np.ndarray:
+def calculate_ifainv(F: np.ndarray, A_inv: np.ndarray) -> np.ndarray:
```
s2t2: As we are updating the formatting of the function names, and some of the parameters (like `ifainv` below), we could consider also updating the formatting of the remaining parameters (like `A_inv` here), mostly just to be consistent in our choices. 

However ultimately it would be great to rename the variables to be more descriptive and less about the mathematical notation. I realize that might be a large lift, so it isn't the biggest priority.

At this time, renaming variables isn't a major concern, but I did want to mention this during my review.

smart_control/simulator/building_radiation_utils.py
```diff
@@ -788,47 +830,73 @@ def mark_directly_seeing_nodes(
...
 
   This function identifies and marks wall nodes that have a direct line of sight
   to the base node. It processes all connected wall nodes (marked with
   marked_value) and determines which ones can directly see the base node without
-  being blocked by other walls.
+  being blocked by other walls. Additionally, it marks air nodes along unblocked
+  lines of sight between wall nodes for interior mass radiative heat transfer.
+
+  When the base node is an air cell, it finds directly seeing nodes among the
+  interior walls, but does NOT mark air nodes as AIR_IN_LINE_OF_SIGHT.
```
s2t2: Let's prefer to use backticks to implement markdown code formatting for variable names like `AIR_IN_LINE_OF_SIGHT` and `constants.INTERIOR_WALL_VALUE_IN_FUNCTION` in docstrings.

Let's also prefer to use quotes when referencing file names like "constants.py" in docstrings.

smart_control/simulator/building_radiation_utils_test.py
```diff
@@ -25,31 +30,31 @@ def test_calculate_A_tilde_inv_and_IFAinv(self):
...
     ])
 
-    A_tilde_inv = utils.calculate_A_tilde_inv(epsilon, F)
-    IFAinv = utils.calculate_IFAinv(F, A_tilde_inv)
-    with self.subTest("A_tilde_inv shape"):
-      self.assertEqual(A_tilde_inv.shape, F.shape)
-    with self.subTest("IFAinv shape"):
-      self.assertEqual(IFAinv.shape, F.shape)
+    a_tilde_inv = utils.calculate_a_tilde_inv(epsilon, F)
+    ifainv = utils.calculate_ifainv(F, a_tilde_inv)
```
s2t2: Follow-up to previous comments about variable naming - our main goal is to be consistent. Here we see `a_tilde_inv` and `ifainv` , however we might consider using a variable name of `ifa_inv` or something, just to be consistent. This would apply to lots of the variables and parameter names.

But again this is not a major concern at this time, just a "nice to have" change.

smart_control/simulator/building_radiation_test.py
```diff
@@ -71,2613 +77,8 @@ def test_interior_radiative_heat_transfer(self):
...
-              0.0,
-          ],
-      ])
-
-      assert_array_almost_equal(
-          np.round(self.building.interior_wall_VF, 5),
-          np.round(expected_interior_wall_vf, 5),
-      )
+      props = self.building.interior_wall_vf
+      assert_data_file_values_equal(props, "expected_interior_wall_vf.csv")
```
s2t2, line 81: Love that we moved these expected values into a CSV file. Makes the tests much easier to read 🙌 ❤️

smart_control/simulator/building_test.py
```diff
@@ -1714,6 +1714,61 @@ def test_stochastic_convection_simulator_shuffle_max_dist(
...
+        zone_map=zone_map,
+        buffer_from_walls=0,
+        interior_mass_properties=interior_mass_properties,
+        include_interior_mass=True,
+    )
+
+    # Check that interior mass is enabled
+    self.assertTrue(b.include_interior_mass)
+
+    # Check that interior mass mask exists and has correct shape
```
s2t2, line 1746: These comments are helpful in explaining the different behaviors we are testing for. Perhaps we can consider further codifying them into separate subtests (nice to have, not strictly necessary).

```
with self.subTest("All interior space values should be 0 (air nodes)"):
  self.assertEqual(np.sum(b.floor_plan[b.interior_mass_mask]), 0)

# etc. etc.
```

smart_control/simulator/simulator.py
```diff
@@ -185,11 +185,67 @@ def _get_edge_cv_temp_estimate(
   def _get_interior_cv_temp_estimate(
```
s2t2, line 185: This documentation is super helpful! I will add private methods to the docs for the `Simulator` class, so we can surface this information, including the formulas.

Commit to be pushed shortly.

smart_control/simulator/simulator.py
```diff
@@ -326,6 +409,147 @@ def update_temperature_estimates(
...
+      self, air_temperature_estimates: np.ndarray
+  ) -> tuple[np.ndarray, float]:
+    r"""Updates interior mass node temperatures based on heat transfer with air
+       CVs.
+
+    Interior mass nodes are adiabatic (no interaction with each other) and only
+    exchange heat with their corresponding air CV. The heat exchange occurs
+    through the vertical direction (height z) of the control volume.
+
+    Equations:
```
s2t2, line 422: Nice. These docs are awesome. Thank you!!

Image

---

Image

---

Image

smart_control/simulator/tf_simulator.py
s2t2: Here is a preview of this method in the docs:

Image

---

Image

smart_control/simulator/simulator_flexible_floor_plan_test.py
```diff
@@ -219,24 +222,38 @@ def _create_small_building(
         conductivity=2.0, heat_capacity=500.0, density=1800.0
     )
     building_exterior_properties = building_py.MaterialProperties(
-        conductivity=0.05, heat_capacity=500.0, density=3000.0
+        conductivity=1.0, heat_capacity=500.0, density=3000.0
```
s2t2, line 225: Is this value change to make the exterior properties more realistic (issue with the value being used previously)?

This question underscores the helpfulness of including some `MaterialProperties` lookup values / list of values for common materials in the docs.

smart_control/simulator/simulator_flexible_floor_plan_test.py
```diff
@@ -275,6 +294,59 @@ def _create_small_building(
 
     return building
 
+  # def _create_small_building_with_interior_mass(
```
s2t2: this method is commented out. let's use or remove it before merging

s2t2: if this helper method will allow us to share lots of setup code used in the tests below, let's definitely prefer to refactor the code below to use it. ideally we can move the helper function to the "simulator/conftest.py" to be available outside of this particular test.

s2t2: Gemini comments:

The following blocks of code are repeated across `test_interior_mass_temperatures_update`, `test_interior_mass_convergence`, and are largely present in `test_interior_mass_affects_heat_transfer` and `test_interior_mass_convergence_with_lwx`.

Suggested Refactoring: _create_simulator_and_building

You can create a helper method within the test class to handle the common initialization logic. This would significantly reduce the test function size and improve maintainability.

```py

def _create_simulator_and_building(
    self,
    initial_temp=292.0,
    include_interior_mass=True,
    include_radiative_heat_transfer=False,
    convergence_threshold=0.001,
    iteration_limit=100,
):
    """Creates a building and simulator instance with shared parameters."""
    weather_controller = mock.create_autospec(
        weather_controller_py.WeatherController
    )
    time_step_sec = 300.0
    hvac = self._create_small_hvac()
    iteration_warning = 10
    start_timestamp = pd.Timestamp("2012-12-21")

    # Building geometry and base properties
    cv_size_cm = 20.0
    floor_height_cm = 300.0
    inside_air_properties = building_py.MaterialProperties(
        conductivity=50.0, heat_capacity=700.0, density=1.0
    )
    inside_wall_properties = building_py.MaterialProperties(
        conductivity=2.0, heat_capacity=500.0, density=1800.0
    )
    building_exterior_properties = building_py.MaterialProperties(
        conductivity=0.05, heat_capacity=500.0, density=3000.0
    )
    interior_mass_properties = building_py.MaterialProperties(
        conductivity=0.5, heat_capacity=1000.0, density=2000.0
    )

    floor_plan = self._create_dummy_floor_plan_small()
    zone_map = copy.deepcopy(floor_plan)

    building = building_py.FloorPlanBasedBuilding(
        cv_size_cm=cv_size_cm,
        floor_height_cm=floor_height_cm,
        initial_temp=initial_temp,
        inside_air_properties=inside_air_properties,
        inside_wall_properties=inside_wall_properties,
        building_exterior_properties=building_exterior_properties,
        floor_plan=floor_plan,
        zone_map=zone_map,
        buffer_from_walls=0,
        interior_mass_properties=interior_mass_properties,
        include_interior_mass=include_interior_mass,
        include_radiative_heat_transfer=include_radiative_heat_transfer,
        view_factor_method="ScriptF" if include_radiative_heat_transfer else None,
    )

    simulator = simulator_py.SimulatorFlexibleGeometries(
        building,
        hvac,
        weather_controller,
        time_step_sec,
        convergence_threshold,
        iteration_limit,
        iteration_warning,
        start_timestamp,
    )
    return simulator, building
```

Example of Refactored Test

Using the new function, the original tests become much cleaner:

```py

def test_interior_mass_convergence(self):
    """Test that simulation with interior mass converges."""
    simulator, _ = self._create_simulator_and_building(
        convergence_threshold=0.001, iteration_limit=100
    )

    # Test convergence with same temperature (should converge quickly)
    converged = simulator.finite_differences_timestep(
        ambient_temperature=292.0, convection_coefficient=12.0
    )

    self.assertTrue(
        converged,
        msg=(
            "Simulation with interior mass should converge when ambient temp"
            " equals initial temp"
        ),
    )
```

smart_control/simulator/simulator_flexible_floor_plan_test.py
```diff
@@ -1562,6 +1636,360 @@ def test_update_temperature_estimates_return_value_with_radiative_heat_transfer_
...
+    # Run a timestep with different ambient temperature to cause heat transfer
+    converged = simulator.finite_differences_timestep(
+        ambient_temperature=300.0, convection_coefficient=12.0
+    )
+
+    self.assertTrue(converged)
+
+    # Check that interior mass temperatures have changed
+    temps_changed = False
+    for x in range(building.interior_mass_mask.shape[0]):
```
s2t2: Gemini comments about `test_interior_mass_temperatures_update` :

Assertion Precision: Instead of the manual loop with a comparison against 1e-10, consider using a NumPy function like `np.any(np.abs(building.interior_mass_temp - initial_interior_mass_temps) > 1e-10)`. This is more idiomatic for array comparisons and makes the intent clearer. You'll need to `import numpy as np`.

Clarity of Check: The manual loop logic is unnecessarily verbose. A simple `np.allclose(building.interior_mass_temp, initial_interior_mass_temps)` negated would be clearer, possibly with an atol argument.

smart_control/simulator/simulator_flexible_floor_plan_test.py
```diff
@@ -1562,6 +1636,360 @@ def test_update_temperature_estimates_return_value_with_radiative_heat_transfer_
...
+        time_step_sec,
+        convergence_threshold,
+        iteration_limit,
+        iteration_warning,
+        start_timestamp,
+    )
+
+    simulator_with_mass = simulator_py.SimulatorFlexibleGeometries(
+        building_with_mass,
+        self._create_small_hvac(),
```
s2t2: do we want to reuse the existing `hvac` variable here instead of initializing a new one / duplicate function call?

smart_control/simulator/simulator_flexible_floor_plan_test.py
```diff
@@ -1562,6 +1636,360 @@ def test_update_temperature_estimates_return_value_with_radiative_heat_transfer_
...
+    # differ
+    self.assertNotAlmostEqual(
+        avg_temp_no_mass,
+        avg_temp_with_mass,
+        places=12,
+        msg=(
+            "Buildings with and without interior mass should have different"
+            " average temperatures"
+        ),
+    )
```
s2t2, line 1756: Gemini comments about `test_interior_mass_affects_heat_transfer`:

Assertion Robustness: The test asserts `assertNotAlmostEqual`. While this confirms a difference, it relies heavily on the specific material properties and time step chosen. A more robust test might check for the expected direction of the change (e.g., that the building with mass heats up slower than the one without, meaning `avg_temp_with_mass < avg_temp_no_mass` in this specific setup).

smart_control/simulator/simulator_flexible_floor_plan_test.py
```diff
@@ -1562,6 +1636,360 @@ def test_update_temperature_estimates_return_value_with_radiative_heat_transfer_
...
+  def test_interior_mass_convergence_with_lwx(self):
+    """Test that simulation with interior mass converges with LWX
+    (longwave interior radiative heat transfer)."""
+    weather_controller = mock.create_autospec(
+        weather_controller_py.WeatherController
+    )
+    time_step_sec = 300.0
+    hvac = self._create_small_hvac()
+    convergence_threshold = 0.01
+    iteration_limit = 1000
```
s2t2: Gemini comments about `test_interior_mass_convergence_with_lwx`:

Iteration Limit/Threshold: The convergence_threshold is $0.01$ and iteration_limit is $1000$. This suggests this scenario might be harder to converge than the others. Add a comment explaining why these values are different, or consider unifying them if possible, unless the use of longwave radiation fundamentally necessitates a different convergence approach.

s2t2 (changes requested):
Hi @ecosang I have completed my initial review. I also shared some comments from Gemini.

This looks great in general!

The main change I would like to see is to refactor some of the shared setup steps for the tests (which I believe you were starting to do). 

Also as previously mentioned, it would be great to add more documentation about common material property values. It would be nice to also include some comments about why we are using certain values in our tests.

Feel free to mark any comments / conversations as resolved if they don't require further action.

smart_control/simulator/simulator.py
s2t2: UPDATE: I am experiencing some SSH connection issues on my work computer that are preventing me from pushing a commit at the moment. Will look into them and push the commit when the issues are resolved.

s2t2: Turns out it was an issue related to a GitHub outage. But it is resolved now and I was able to push those commits.
https://www.howtogeek.com/github-is-down-its-not-just-you/

smart_control/simulator/conftest.py
ecosang: @s2t2 Hi Michael, I will explain during the meeting, but this is not a straightforward process.  The main reason is sbsim uses the same properties for all wall nodes, which means if there are multiple layers of insulation, it is hard to define them. Another case is that the indoor air property is somewhat unrealistic. The original implementation has the indoor as somewhat higher mass to capture interior_mass property to some extent. Also, indoor air heat transfer is driven by mixing/convection, while sbsim models it as a conduction process, which is not true.  I think this is because the sbsim is targeting data-driven calibration instead of specifying material properties. So, if we capture the behavior with some number to some extent from the data-driven calibration, I think it is fine. To sum, the material properties are not really physical numbers, but lumped values to capture thermal behaviors. 
Radiative one is different. We are dealing with the material directly (surfaces), so we can define it still in sbsim framework.

 The current default values are reasonable in my opinion.

"""
inside_air_properties = building.MaterialProperties(
conductivity=50.0, heat_capacity=700.0, density=1.2
)
inside_wall_properties = building.MaterialProperties(
conductivity=2.0, heat_capacity=1000.0, density=1800.0
)
building_exterior_properties = building.MaterialProperties(
conductivity=0.05, heat_capacity=1000.0, density=3000.0
)
"""

smart_control/simulator/simulator_flexible_floor_plan_test.py
ecosang: I've made changes. Originally, the air mass is very high, and it takes lots of time to make a convergence when there are interior mass especially, especially when the initial temperature and outdoor temperature are very different. But, after putting realistic air mass value, this issue disappears. 
It is specifically a concern when doing the first step of finite difference because we need several iterations from the initial values. However, during the simulation periods, it will be fine as now we only consider the temperature changes from the last time step.

smart_control/simulator/simulator_flexible_floor_plan_test.py
ecosang: I replaced it with other function to create a simulator.

smart_control/simulator/simulator_flexible_floor_plan_test.py
ecosang: this is revised.

smart_control/simulator/building_radiation_utils_test.py
ecosang: updated to ifa_inv

smart_control/simulator/building_radiation_utils.py
ecosang: updated to ifa_inv.

smart_control/simulator/building_radiation_utils.py
ecosang: I updated in [d8354c1](https://github.com/google/sbsim/pull/131/commits/d8354c15962ec398608d4e63942f0b93dcd1dd36)

smart_control/simulator/building_test.py
ecosang: [b9225d0](https://github.com/google/sbsim/pull/131/commits/b9225d005417eda726e91ce3144831c87507d1ff)

smart_control/simulator/simulator_flexible_floor_plan_test.py
ecosang: I have refactored.. [d8354c1](https://github.com/google/sbsim/pull/131/commits/d8354c15962ec398608d4e63942f0b93dcd1dd36)

smart_control/simulator/simulator_flexible_floor_plan_test.py
ecosang: [d8354c1](https://github.com/google/sbsim/pull/131/commits/d8354c15962ec398608d4e63942f0b93dcd1dd36)

smart_control/simulator/simulator_flexible_floor_plan_test.py
ecosang: [4cae776](https://github.com/google/sbsim/pull/131/commits/4cae776cbea5e836d84f9cd304177d1bf4bdba12)

smart_control/simulator/conftest.py
s2t2: @ecosang thanks for the explanation. 

Quick follow-up question: If I am understanding correctly (that some of the currently used material property values are unrealistic because they are compensating for a lack of de-tangling between inside air vs inside mass), would we then want to use different default values depending on whether or not the interior mass mode is on? For example we can choose more realistic values if interior mass mode is on.

Let me know if this suggestion is reasonable and helpful, or way off base 😅

ecosang: @s2t2 Yes, your interpretation is correct.
For the exterior and interior walls, I think I can do some calculations and provide some default values. 

For example, a typical wall composition's R value is 2.5 (m2K/W) and its thckness is 0.13m, so the averaged conductivity is 0.052 W/mK. But, for the capacity and density, I need to find all the materials information and wall composition (thickness) and need to calculate the weighted average value.

For the interior wall, I can do something similar. 

Which is doable, but the results are similar to the current default values.


The air node is the problem. When using interior mass, yes, we can put a realistic number. Without interior_mass, we cannot. We can use some numbers that incorporate interior_mass+air information altogether, like the current implementation. 

After all, the sbsim's core work is data-driven model, and the calibration of material properties is essential. 
Based on the floor plan and measured data, the material properties need to be tuned. So, the default values are not very useful in my view. On the other hand, the radiative heat transfer is slightly different. Their impacts might be minimal in the simulation, and using default values are more robust as they are a direct material property values.

At this point, I have several things to do, so I would like to leave it as it is, and once I have more confidence in the interior_mass model, I would like to have default values. (Another thing is, we don't have inter-air heat transfer through mixing in sbsim.  sbsim overcomes this by having high conductivity number for the air, but the air's conductivity is really small. Instead, the inter-air heat transfer is achieved via air mixing, not conduction. With the interior_mass model, if I put a realistic number for air (i.e., low conductivity), then the inter-air heat transfer may disappear. So we need to think of how to model this..)

ecosang:
@s2t2 I resolved all your comments in my opinion, except for the default values.. Do you have any other suggestions? It is not urgent, but I want to make a new branch based on here to continue exterior radiative heat transfer modeling.

smart_control/simulator/conftest.py
s2t2: OK, thank you for the explanation. Sounds good.

Let me know if you plan on making any other updates to this PR, or if you think it is ready for final review and merge.

ecosang: @s2t2 I am good with this PR. Other updates come in separate PRs.

s2t2 (approved):
LGTM  🙌 🚀

PR #133, opened 2025-12-14
Fix Intel-based macOS build

### LINKED ISSUE #126: Troubleshoot and Fix the Mac Build
We see that the [build](https://github.com/google/sbsim/blob/copybara_push/.github/workflows/build.yaml) is passing on Linux, but [failing on Mac](https://github.com/google/sbsim/actions/runs/18141287523/job/51632482976). 

It could be that things are just timing out, and things are taking too long to install via homebrew. We noticed the Mac build was taking a very long time. But it could be a different issue. 

We need to troubleshoot the issue, and ideally fix it to keep Mac support. Otherwise we should remove Mac support from the build (not ideal, only if necessary).

Someone who is comfortable with GitHub Actions would be great person to tackle this issue!

AvishKaushik:
## Summary
Fixes the Intel-based macOS build failures by upgrading from the deprecated `macos-13` runner to the new `macos-15-intel` runner.

## Problem
The Intel-based macOS build was disabled due to build failures. Investigation revealed that the `macos-13` runner has been deprecated by GitHub Actions and is no longer receiving updates, causing the build to fail.

## Solution
- Updated the Intel Mac runner from `macos-13` to `macos-15-intel`
- Re-enabled the Intel-based macOS build that was temporarily disabled
- `macos-15-intel` is the official replacement for Intel-based macOS builds and will be supported until August 2027

## Testing
All CI builds are now passing:
- Ubuntu (Python 3.10.15, 3.11)
- macOS ARM (`macos-latest`)  
- macOS Intel (`macos-15-intel`)

Test run: https://github.com/AvishKaushik/sbsim/actions/runs/20201182727

## References
- [GitHub Actions: macOS 13 runner image is closing down](https://github.blog/changelog/2025-09-19-github-actions-macos-13-runner-image-is-closing-down/)
- [macOS 15 Intel-based image availability](https://github.com/actions/runner-images/issues/13045)

Resolves #126

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -22,12 +22,8 @@ jobs:
         python-version: ['3.10.15', '3.11']
 
         include:
-          # temporarily disabling this Intel-based Mac OS run
-          # because of issues with underlying dependencies breaking the build
-          # see: https://github.com/google/sbsim/issues/126
-          #
-          #- os: macos-13 # intel-based
-          #  python-version: '3.11'
+          - os: macos-15-intel # updated intel-based
+            python-version: '3.11'
 
           - os: macos-latest # arm-based
             python-version: '3.11'


s2t2:
@AvishKaushik thank you so much!! The Intel Mac build is back! 💪

Issue #144, opened 2026-01-16
Weather Data Cleaning


s2t2:
> NOTE: this issue references a future state of the code that has not yet been synced out to github

We see the weather data uses extreme values like -9999.0 to signal a null or bad value. For example in the "TempC" column.

To avoid downstream issues when using this data in simulation, we should model those values as null or nan instead. 

We should then also make sure the Replay Weather Controller appropriately handles / drops those null values (instead of using a very extreme negative number). Using the interpolation strategy, the absence of one hour won't negatively impact the results because the weather controller will use the average between the closest two hours it does have.

For the "TempC" column, the following dates have a single instance of these bad / extreme values:

2025:
  + 04-02
  + 06-06
  + 07-07
   + and possibly others not included since the current data snapshot ends before the end of the year

2024:

  + 06-04
  + 07-04
  + 08-03
  + 11-16
  + 11-24
  + 12-24

2023:

  + 04-20

As a temporary workaround we are performing the filtering downstream:

```py
wc = ReplayWeatherController()
# or access from the environment:
# wc = env.building.simulator.weather_controller
wc.weather_df = wc.weather_df[ wc.weather_df["TempC"] != -9999.0]
# now you can use the weather controller
```

s2t2:
FYI here are the temp ranges per week in each dataset (2025 not complete and needs to be re-pulled)

<img width="843" height="524" alt="Image" src="https://github.com/user-attachments/assets/16e70b6b-f59c-417d-b527-953187e6b60a" />

<img width="843" height="524" alt="Image" src="https://github.com/user-attachments/assets/46b33509-d361-4c3b-91af-18d6a61f8d1f" />

<img width="843" height="524" alt="Image" src="https://github.com/user-attachments/assets/2ea45171-c340-4f17-9213-f7db7735b492" />

s2t2:
This is being fixed internally by @amcberkes.

Oh also we are interpolating missing values instead of dropping them 🙌 🏆 💪 

We will mark this issue closed once the fix is synced out and merged.

PR #146, opened 2026-02-06
[LBNL] Exterior radiative heat transfer implementation

ecosang:
- Fix interior radiative heat transfer errors (missing to multiply area)
- Add solar radiation calculation in weather controller (and several utility functions in building_radiation_utils.py)
- Add exterior radiative heat transfer related utilities
- Add fenestration constants
- Add docstring for heat balance calculation for simulator.py
- Add tensorized exterior radiative heat transfer
- Add tensorized exterior radiative heat transfer test cases (with validation with iterative approach + visualization+speed benchmakr).
- Add relative_max_delta change metric for simulation convergence (so if max_delta doesn't change for x iteration for <threshold, the iteration stop). The main reason is when we have interior_mass nodes, the interior_mass node's temperature change becomes very small after several iteration. We can consider it as a convergence.
@s2t2 please assign the reviewers.


s2t2:
@ecosang nice! I will aim to review early next week

smart_control/simulator/building_radiation_test_data/723170TYA.CSV
s2t2, line 1: The units in the column names are helpful.

I wasn't sure about the order of the rows, but perhaps it would be helpful to sort them on the basis of timestamp ascending. This would help provide a better understanding of the time range.

smart_control/simulator/benchmark_results_23x12.txt
s2t2, line 1: Because these files aren't necessarily code, let's consider moving them to the "docs" directory:
  + smart_control/simulator/benchmark_results_23x12.txt
  + smart_control/simulator/temperature_comparison_lwx_lwr.png

Perhaps we can add them to a new "docs/experiments/YOUR_EXPERIMENT_NAME_HERE" directory.

In that directory, you can feel free to create a new docs page (like "index.md") to display these results (and eventually link to the paper).

smart_control/simulator/weather_controller.py
```diff
@@ -48,11 +61,64 @@ def __init__(
       default_high_temp: float,
       special_days: Optional[Mapping[int, TemperatureBounds]] = None,
       convection_coefficient: float = 12.0,
+      tz: str = 'UTC',
```
s2t2: Let's prefer the parameter name `time_zone`, to be consistent with other implementations across the codebase.

smart_control/simulator/weather_controller.py
```diff
@@ -113,10 +196,182 @@ def get_air_convection_coefficient(self, timestamp: pd.Timestamp) -> float:
...
+    return {
+        'ghi': max(0, ghi),
+        'dni': max(0, dni),
+        'dhi': max(0, dhi),
+        'solar_zenith': solar_zenith,
+        'solar_azimuth': solar_azimuth,
+    }
+
+  def get_current_sky_temperature(self, timestamp: pd.Timestamp) -> float:
+    """Returns sky temperature in K using Clark & Allen formula.
```
s2t2, line 399: It would be great to provide links to information about the methods or formula used. 

This comment applies more broadly, to other classes and functions, as applicable. 

I do see source links in some places, but if we can expand the coverage it would be helpful.

s2t2 (commented):
I started providing a few comments, but then I noticed this is a very large PR (~15K lines of code). I want to give a thorough review, but the large size is presenting a challenge for my own comprehension and attention. The large size may also present a challenge when we try to sync this PR into the Google codebase - internally the rule of thumb is to try to keep PRs under ~1K lines, if possible (although this is not a strict requirement).

So I would like to see us try to break this up into multiple smaller PRs, to facilitate a high quality, detailed review process. For example, perhaps we could do one for the weather controller, one for the updated radiation utils, etc.

Let me know how you would like to plan on splitting up the functionality, and the rough order we would want to review it in. What would you say is the chain of dependencies here? If you consider what functionality depends on others, we would probably want to start with functionality that has no dependencies, that other functionality depends on. 

We can keep this large PR open for now, for reference, and close it once all the individual components have been merged separately.

smart_control/simulator/weather_controller.py
s2t2, line 1: FYI internally we have refactored the Replay Weather Controller a bit, to use a shared `_get_interpolated_value` method that can get the value for any column at that point in time.

You may want to use this updated code as the basis for any changes / updates to the Replay Weather Controller.

I am going to try to create a quick PR to update the `copybara_push` branch to have this updated content for you to work on top of. I would include the updated tests as well. Will report back here with updates.

UPDATED WEATHER CONTROLLER FILE CONTENTS:

FYI -  currently in the `_get_interpolated_value` function, I am passing the list of values to be interpolated, but this requires a separate property for each column of values (e.g. `humidities`). A more flexible interface would have us pass the column name (instead of the values) into the `_get_interpolated_value` function.

```py

# ... 

@gin.configurable
class ReplayWeatherController(BaseWeatherController):

  def __init__(
      self,
      local_weather_path: str = WEATHER_CSV_FILEPATH,
      convection_coefficient: float = 12.0,
      humidity_column: str = 'Humidity',
  ):
    self.local_weather_path = local_weather_path
    self.weather_df = self.read_weather_csv(self.local_weather_path)
    self.convection_coefficient = convection_coefficient
    self.humidity_column = humidity_column

  @property
  def csv_filepath(self) -> str:
    """Alias for the local weather CSV file path."""
    return self.local_weather_path

  def read_weather_csv(self, csv_filepath: str) -> pd.DataFrame:
    df = pd.read_csv(csv_filepath)
    df = df.drop(columns=['Unnamed: 0'], errors='ignore')

    df['Time'] = pd.to_datetime(df['Time'], utc=True)

    df.index = (df['Time'] - _EPOCH).dt.total_seconds()
    df.index.name = 'SecondsSinceEpoch'

    return df

  @property
  def min_time(self) -> pd.Timestamp:
    return min(self.weather_df['Time'])

  @property
  def max_time(self) -> pd.Timestamp:
    return max(self.weather_df['Time'])

  @property
  def times_in_seconds(self) -> pd.Index:
    return self.weather_df.index

  @property
  def temps_f(self) -> pd.Series:
    return self.weather_df['TempF']

  @property
  def humidities(self) -> pd.Series:
    return self.weather_df[self.humidity_column]

  def _get_interpolated_value(
      self, timestamp: pd.Timestamp, values: pd.Series
  ) -> float:
    """Helper to get interpolated value from a given series.

    The timestamp need not exactly appear in the weather data, but should be
    within the range of the data.
    If there is no exact match, linear interpolation is used to estimate the
    temperature between the nearest timestamps.

    Args:
      timestamp: Pandas timestamp to get temperature for interpolation. If the
        timestamp is timezone aware, it will be converted to UTC. If the
        timestamp is timezone naive, it will be localized to UTC. This allows
        for accurate comparisons against the min and max timestamps, as well as
        the epoch, which are always timezone aware (in UTC).
      values: Pandas series to interpolate from.

    Returns:
      The interpolated value from the series at the given timestamp.
    """
    # convert timestamp to UTC to enable proper comparisons:
    if timestamp.tzname() is not None:
      # timestamp is timezone aware, unable to localize, so convert to UTC:
      timestamp = timestamp.tz_convert('UTC')
    else:
      # timestamp is timezone naive, unable to convert, so localize to UTC:
      timestamp = timestamp.tz_localize('UTC')

    if timestamp < self.min_time:
      raise ValueError(
          f'Timestamp not in range. Timestamp {timestamp} is before the'
          f' earliest timestamp {self.min_time}.'
      )
    if timestamp > self.max_time:
      raise ValueError(
          f'Timestamp not in range. Timestamp {timestamp} is after the'
          f' latest timestamp {self.max_time}.'
      )

    time_in_seconds = (timestamp - _EPOCH).total_seconds()
    return np.interp(time_in_seconds, self.times_in_seconds, values)

  def get_current_temp(self, timestamp: pd.Timestamp) -> float:
    """For a given timestamp, returns the current temperature in Kelvin."""
    return utils.fahrenheit_to_kelvin(
        self._get_interpolated_value(timestamp, self.temps_f)
    )

  def get_current_humidity(self, timestamp: pd.Timestamp) -> float:
    """For a given timestamp, returns the current humidity level in percent."""
    return self._get_interpolated_value(timestamp, self.humidities)

  # pylint: disable=unused-argument
  def get_air_convection_coefficient(self, timestamp: pd.Timestamp) -> float:
    return self.convection_coefficient
```

s2t2, line 1: Also, I noticed some [buggy behavior](https://github.com/google/sbsim/issues/144) in the Replay Weather Controller when the CSV data file values are -9999. Unfortunately the data provider uses -9999 instead of Null values. 

We should consider cleaning up the data frame, to convert the -9999 to nulls, or filter them out (because the linear interpolation will not be significantly affected). For the TempC -9999 values, I was considering filtering them out. However there are other columns where the values are all or mostly -9999, in which case we should consider using NULL values instead. 

My initial thoughts about strategy for just the TempC values is below. But now you are using lots of other columns, so I welcome your thoughts about how to handle nulls.

```py

  def read_weather_csv(self, csv_filepath: str) -> pd.DataFrame:
    # ...

    df = pd.read_csv(csv_filepath)
    df = df.drop(columns=['Unnamed: 0'], errors='ignore')

    df['Time'] = pd.to_datetime(df['Time'], utc=True)

    df.index = (df['Time'] - _EPOCH).dt.total_seconds()
    df.index.name = 'SecondsSinceEpoch'

    # NEW:
    # filter out known issue with the data where some temps are "null":
    # df = df[df['TempC'] != -9999.0]

    return df
```

ecosang:
@s2t2 It seems like I can put a reply in each thread. Yeah, originally, my intention was to update the weather controller first and the radiation part later in a separate PR. But at some point, I have some minor edits on both while implementing weather controller, so the PR is rather being large. 

Yes, we can do the following steps.
1. Weather controller 
- smart_control/simulator/building_radiation_test_data/723170TYA.CSV
- smart_control/simulator/weather_controller.py
- smart_control/simulator/local_weather_test_data.csv

2. Building radiation utils
- smart_control/simulator/building_radiation_utils.py
- smart_control/simulator/building_radiation_utils_test.py


3. Building related

- smart_control/simulator/building.py
- smart_control/simulator/constants.py
- smart_control/simulator/building_test.py

4. simulator
- smart_control/simulator/simulator.py
- smart_control/simulator/simulator_flexible_floor_plan.py
- smart_control/simulator/simulator_flexible_floor_plan_test.py

5. Tensor
- smart_control/simulator/tf_simulator_test.py
- smart_control/simulator/tf_simulator.py

smart_control/simulator/weather_controller.py
s2t2: Oh, I believe the time zone may always be UTC, in the replay weather controller at least, because of the nature of the CSV data. If that is the case, parameterizing it may not be appropriate, but we can make a static property instead, unless there is a reason we want to customize (apologies I haven't yet gotten into a lot of the details of the newly suggested code).

s2t2:
OK, thanks, sounds like a plan! Let me work on pushing those weather controller updates first, and then I will give you the go ahead.

s2t2:
@ecosang here is the PR with the updated weather controller, including additional years of weather data:
https://github.com/google/sbsim/pull/148

You should now be good to go with Part 1 (Weather Controller).

FYI - Before we get to Parts 3 and 4, I have some minor updates to sync-out for the building and simulator. Mostly just adding a few properties. I will prepare another sync-out sometime over the next few days, and let you know when it is complete.

s2t2:
> @ecosang here is the PR with the updated weather controller, including additional years of weather data: #148
> 
> You should now be good to go with Part 1 (Weather Controller).
> 
> FYI - Before we get to Parts 3 and 4, I have some minor updates to sync-out for the building and simulator. Mostly just adding a few properties. I will prepare another sync-out sometime over the next few days, and let you know when it is complete.

Actually @ecosang it looks like some of those updates are polluted with new sim HVAC updates (to use multiple air handlers), so I think it would be simpler for you to just merge your code in first, and then we will sync-out the HVAC updates afterwards.

ecosang:
HI Michael, I merged the weather part. Can you guide me on how to proceed? Is it possible to split the commits into some pieces.?

s2t2:
> HI Michael, I merged the weather part. Can you guide me on how to proceed? Is it possible to split the commits into some pieces.?

Nice. Thank you. Based on the [assessment of dependencies](https://github.com/google/sbsim/pull/146#issuecomment-3873635531), it seems we might be able to split up the logic at the file level.

I would recommend making a new branch based on the current `copybara_push` branch, and then copying over the specific files from this branch (`ext_radiation`) into that new branch. 

Maybe something like this? Let me know how it goes!

```sh
# ensure your fork is up to date (assuming Google's repo is upstream and your repo is origin):
git checkout copybara_push
git pull upstream copybara_push
git push origin copybara_push

# make a clean new branch (for example called "external-rad-part-1-weather"):
git checkout -b external-rad-part-1-weather

# update the contents of specific files, using their contents from this branch:
# git checkout <source-branch-name> -- <path/to/file>
git checkout ext_radiation -- smart_control/simulator/building_radiation_test_data/723170TYA.CSV
git checkout ext_radiation -- smart_control/simulator/weather_controller.py
git checkout ext_radiation -- smart_control/simulator/local_weather_test_data.csv
```

s2t2:
Regarding the failing CI build, it looks like it can't find some of the weather data files. I will look into this. I may need to include some additional files and/or update the filepaths.

UPDATE: it looks like the data file is present in the repo in the simulator directory: "smart_control/simulator/local_weather_test_data.csv" but the filepath being referenced in the failing build is pointing at the "configs/resources/sb1" directory. We might just need to update the filepath reference to point to the right path to the existing file in the simulator dir. Alternatively if you want to move the CSV file to the resources directory, that would be ok too.

PR #148, opened 2026-02-09
Weather Data Sync-out

s2t2:
Exporting content from internal Google codebase.

Details:

  + Expands coverage of weather data to include additional years (2023, 2024, partial 2025).

  + Refactors the Replay Weather Controller to inherit from the Base Weather Controller, handle humidity, and read from any of the weather data files.


smart_control/simulator/weather_controller.py
```diff
@@ -149,55 +161,138 @@ def get_outside_air_temp(observation_response):
...
+    """Returns the temperatures in Fahrenheit of the weather data."""
+    return self.weather_df['TempF']
+
+  @property
+  def humidities(self) -> pd.Series:
+    """Returns the humidities of the weather data."""
+    return self.weather_df[self.humidity_column]
+
+  def _get_interpolated_value(
+      self, timestamp: pd.Timestamp, values: pd.Series
```
s2t2, line 245: We are currently passing the list of values to be interpolated, but this requires a separate property for each column of values (e.g. `humidities`, etc.). 

If possible, let's implement a more flexible interface where we can just pass the column name (instead of the values) into the `_get_interpolated_value` function.

smart_control/configs/resources/sb1/weather_data/2025.csv
s2t2, line 1: Contains data up until 8/25. We need to update the data to include the full year.

We have a pipeline for updating this data, so this issue should be handled internally by Google staff.

smart_control/simulator/weather_controller.py
```diff
@@ -149,55 +161,138 @@ def get_outside_air_temp(observation_response):
...
+    Args:
+      csv_filepath: Path to local weather CSV file.
+
+    Returns:
+      Pandas dataframe of weather data.
+    """
+    df = pd.read_csv(csv_filepath)
+    df = df.drop(columns=['Unnamed: 0'], errors='ignore')
+
+    df['Time'] = pd.to_datetime(df['Time'], utc=True)
```
s2t2, line 212: Consider cleaning the data here. See https://github.com/google/sbsim/issues/144

s2t2 (commented):
OK to merge. Adding some issues which can be addressed separately.

PR #165, opened 2026-03-16
LBNL - Solar Radiation

ecosang:
add solar calculation in weather_controller and define fenestration constants


ecosang:
@s2t2 I've put only weather controller part in this PR.

s2t2:
OK, thanks! I see the build is failing due to CSV path issues (see below). Would you please fix and then ping me and I will review again when the build is passing.

> FAILED smart_control/simulator/weather_controller_test.py::ReplayWeatherControllerPvlibValidationTest::test_replay_controller_irradiance_closure_equation - FileNotFoundError: [Errno 2] No such file or directory: '/home/runner/work/sbsim/sbsim/smart_control/simulator/../configs/resources/sb1/weather_data/local_weather_test_data.csv'

ecosang:
@s2t2 the csv file is automatically ignored. I just added. Sorry for the iteration.  Can I also check if the testing is passed here on my end? when I made PR, I didn't see the failure. Maybe I didn't wait enough.

s2t2:
For some reason, I have to click the "Approve workflows and run" button to trigger the tests. Running now...

smart_control/simulator/local_weather_test_data.csv
s2t2, line 1: There is an existing test data file in "smart_control/simulator/local_weather_test_data.csv".  I was wondering how these files differ, and if we could leverage the existing file. Or organize both of these files to be in the same place (perhaps putting them in a "smart_control/simulator/test_data" folder.)

smart_control/simulator/solar_radiation_test_data/723170TYA.CSV
s2t2, line 1: It would be nice to know how this data was acquired, what the source is, and perhaps also to have access to the code that pulled this data (as applicable).

smart_control/simulator/local_weather_test_data.csv
s2t2: Oh I see you moved this file from that location, to be in the resources dir. That could be OK. However I would prefer to keep both test data files in the same location.

smart_control/simulator/building_radiation_utils.py
```diff
@@ -967,3 +967,47 @@ def mark_directly_seeing_nodes(
...
+      and Diffuse Horizontal Irradiance in W/m2.
+    surface_tilt: Surface tilt angle from horizontal in degrees (0 = horizontal,
+      90 = vertical).
+    surface_azimuth: Surface azimuth angle in degrees (180 = south-facing in
+      Northern Hemisphere, compass direction that the surface normal points).
+    solar_zenith: Solar zenith angle in degrees (angle from vertical).
+    solar_azimuth: Solar azimuth angle in degrees (compass direction of sun).
+  Returns:
+    POA global irradiance in W/m2.
+  Example:
```
s2t2: Appreciate the example in the docstring.

smart_control/simulator/building_radiation_utils.py
```diff
@@ -967,3 +967,47 @@ def mark_directly_seeing_nodes(
...
+    >>> irrad = {'ghi': 800.0, 'dni': 700.0, 'dhi': 100.0}
+    >>> poa = calculate_poa_irradiance(irrad, surface_tilt=30.0,
+    ...     surface_azimuth=180.0, solar_zenith=30.0, solar_azimuth=180.0)
+  """
+  # Import here to avoid circular imports and keep pvlib as optional dependency
+  from pvlib import irradiance as pvlib_irradiance  # pylint: disable=import-outside-toplevel
```
s2t2: Can you please further explain the circular dependency concerns if this import were placed at the top of the file? And also separately some reasons we would want to keep pvlib as an optional dependency?

FYI there is a way we can do [optional dependencies using pyproject.toml](https://github.com/google/sbsim/blob/copybara_push/pyproject.toml#L52C1-L53C1), like for example to great a new group called "solar": 

```
[tool.poetry.group.docs.solar]
pvlib = "0.13.1"
```

Then it can be installed like `poetry install --with solar` (see Setup docs related to [notebook setup](https://google.github.io/sbsim/setup/linux/#notebook-setup), for an example.

smart_control/simulator/building_radiation_utils.py
```diff
@@ -967,3 +967,47 @@ def mark_directly_seeing_nodes(
...
+    surface_tilt: float,
+    surface_azimuth: float,
+    solar_zenith: float,
+    solar_azimuth: float,
+) -> float:
+  """Calculate plane-of-array (POA) global irradiance.
+  Converts horizontal irradiance components (GHI, DNI, DHI) to the irradiance
+  incident on a tilted surface. Uses the pvlib library's get_total_irradiance
+  function.
+  Args:
```
s2t2: Nit: I believe we might need a new line before each of these sections in the docstring: Args, Returns, Example. And I believe Return should go last.

smart_control/simulator/weather_controller_test.py
```diff
@@ -151,17 +176,1324 @@ def test_replay_weather_controller_raises_error_after_range(self):
...
+        latitude=latitude,
+        longitude=longitude,
+        timezone='America/Los_Angeles',
+    )
+
+    # Test at noon on a summer day
+    timestamp = pd.Timestamp('2023-07-01 12:00:00', tz='America/Los_Angeles')
+    irrad = weather.get_current_irradiance(timestamp)
+
+    # Check that all components are present and positive
```
s2t2: Appreciate these comments in the test function. 

Let's prefer to use a subTest for each of these blocks of related assertions. Can keep the comment above the sub-test, or as the first indented line within the sub-test.

```
# comment here
with self.subTest(name="______"):
  # or comment here
  # assertions ...
  # more assertions,
  # etc
``

File: smart_control/simulator/weather_controller_test.py [hunk 5]
```diff
@@ -151,17 +176,1324 @@ def test_replay_weather_controller_raises_error_after_range(self):
...
+  def test_get_current_irradiance_no_location(self):
+    """Test that irradiance calculation raises error without location."""
+    low_temp = 273.15
+    high_temp = 298.15
+
+    weather = weather_controller.WeatherController(low_temp, high_temp)
+
+    timestamp = pd.Timestamp('2023-07-01 12:00:00', tz='UTC')
+
+    with self.assertRaises(ValueError):
```
s2t2: For asserting errors, Let's prefer to use `self.assertRaisesRegex`, denoting the error message in the assertion as well.

https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaisesRegex

smart_control/simulator/weather_controller_test.py
```diff
@@ -151,17 +176,1324 @@ def test_replay_weather_controller_raises_error_after_range(self):
...
+      weather.get_current_irradiance(timestamp)
+
+  def test_get_irradiance_with_solar_position_weather_controller(self):
+    """Test irradiance with solar position for WeatherController."""
+    low_temp = 273.15
+    high_temp = 298.15
+    latitude = 37.4
+    longitude = -122.1
+
+    weather = weather_controller.WeatherController(
```
s2t2: If we have a few tests with a shared set up, it might be worthwhile to organize them in their own test class.

For example

```
class WeatherControllerRadiationTest:

  def setUp(self):
    super().setUp()
    self.low_temp = 273.15
    # etc etc
    self.weather = WeatherController(...)
```

This might help simplify the tests and reduce the number of lines of test code, if we find opportunities to refactor the shared setup methods.

In this case because we have so many tests related to radiation, I would suggest organizing them in separate test classes. I would suggest doing one radiation test class for the weather controller, and another for the replay weather controller, and inheriting the replay weather controller radiation test from the weather controller radiation test, to the extent it makes sense.

Also, FYI - we will be undergoing an effort to separate the weather controller and replay weather controller (and their corresponding tests) into separate files. So to not proliferate their coupling, I might suggest creating two separate test files to put these new test classes in (one for weather controller radiation tests and the other for replay weather controller radiation tests). We can leave existing tests where they are for now.

smart_control/simulator/weather_controller.py
```diff
@@ -30,6 +34,47 @@
...
+  """
+  parts = dms_str.strip().split()
+  degrees = float(parts[0])
+  minutes = float(parts[1]) if len(parts) > 1 else 0.0
+  seconds = float(parts[2]) if len(parts) > 2 else 0.0
+  sign = -1 if degrees < 0 else 1
+  return degrees + sign * (minutes / 60.0 + seconds / 3600.0)
+
+
+def load_station_info(station_json_path: str) -> tuple[float, float, str]:
```
s2t2: Nice, I'm glad we can make use of that station JSON file.

I might suggest an approach like this, with a dedicated class and class constructor to initialize an instance from a JSON filepath:

```
# perhaps makes sense for this to be a dataclass
@dataclasses.dataclass
class WeatherStation:
  def __init__(self, lat, lon, time_zone):  # can add default values if desired
    self.lat = lat
    # etc

  @classmethod
  def from_json_file(cls, json_filepath:str):
    data = # read the JSON file
    return cls(lat=parse_dms(data["lat"]), lon=parse_dms(data["lng"]), time_zone=data["timezone"])
```

smart_control/simulator/weather_controller.py
```diff
@@ -160,28 +458,358 @@ def get_outside_air_temp(observation_response):
...
       convection_coefficient: float = 12.0,
+      station_json_path: str | None = None,
+      timezone: str | None = None,
+      latitude: float | None = None,
+      longitude: float | None = None,
+      irradiance_method: str = 'campbell_norman',
       humidity_column: str = 'Humidity',
   ):
     self.local_weather_path = local_weather_path
-    self.weather_df = self.read_weather_csv(self.local_weather_path)
```
s2t2, line 181: I would like to keep this weather_df attribute. Looks like it got renamed, but I would prefer to keep the name as `weather_df` to denote it is a dataframe.

smart_control/simulator/weather_controller.py
```diff
@@ -160,28 +458,358 @@ def get_outside_air_temp(observation_response):
...
     self.convection_coefficient = convection_coefficient
     self.humidity_column = humidity_column
 
+    # Load lat/lon/timezone from station JSON if provided; explicit args
+    # override the values loaded from JSON.
+    if station_json_path is not None:
+      station_lat, station_lon, station_tz = load_station_info(
```
s2t2: Maybe we can pass an instance of the `WeatherStation` class I suggested, and use that here.

This would allow us to remove the following params from the initialization:

  + station_json_path: str | None = None,
   + timezone: str | None = None,
  +  latitude: float | None = None,
   + longitude: float | None = None,

In favor of:

```py
self.latitude = self.station.lat
self.longitude = self.station.lon
self.time_zone = self.station.time_zone
```

smart_control/simulator/weather_controller.py
```diff
@@ -60,11 +116,64 @@ def __init__(
...
+    self.latitude = latitude
+    self.longitude = longitude
+    self.dewpoint_depression = dewpoint_depression  # Dry bulb - dew point (K)
+    self.cloud_cover = cloud_cover
+    self.cloud_cover_low = cloud_cover_low
+    self.cloud_cover_high = cloud_cover_high
+    self.irradiance_method = irradiance_method
+
+    # Validate cloud_cover (static mode)
+    if self.cloud_cover is not None:
```
s2t2: Let's consider moving this cloud cover validation section (lines 142 to 160) into a protected validation method, that we could invoke in the init method like `self._validate_cloud_cover(self.cloud_cover_low, self.cloud_cover_high)`

smart_control/simulator/weather_controller.py
```diff
@@ -60,11 +116,64 @@ def __init__(
...
+        raise ValueError('cloud_cover_low must be between 0 and 100.')
+      if self.cloud_cover_high < 0 or self.cloud_cover_high > 100:
+        raise ValueError('cloud_cover_high must be between 0 and 100.')
+      if self.cloud_cover_low > self.cloud_cover_high:
+        raise ValueError(
+            'cloud_cover_low cannot be greater than cloud_cover_high.'
+        )
+
+    # Validate irradiance_method
+    valid_methods = ('clearsky', 'linear', 'campbell_norman')
```
s2t2: Let's prefer to define these constants at the top of the file, and also document the differences between these methods (perhaps in the weather controller Docstring)

smart_control/simulator/weather_controller.py
```diff
@@ -60,11 +116,64 @@ def __init__(
...
+    if self.irradiance_method not in valid_methods:
```
s2t2: we could consider moving this validation logic to a private method as well, although it isn't as large as the cloud cover validation

smart_control/simulator/weather_controller.py
```diff
@@ -296,4 +933,165 @@ def get_current_humidity(self, timestamp: pd.Timestamp) -> float:
...
+        (no timezone), will be localized to the controller's timezone.
+
+    Returns:
+      Sky temperature in K.
+
+    Raises:
+      ValueError: If timestamp is outside weather data range or sky temperature
+      not calculated.
+    """
+    if 'TempSkyC' not in self._weather_data:
```
s2t2: Let's consider making all the specific important column names into constants, like `SKY_TEMP_COL="'TempSkyC'"` or `SKY_TEMPERATURE_COLUMN_NAME="TempSkyC"`

smart_control/simulator/weather_controller.py
```diff
@@ -296,4 +933,165 @@ def get_current_humidity(self, timestamp: pd.Timestamp) -> float:
...
+        'ghi' not in self._weather_data
+        or 'dni' not in self._weather_data
+        or 'dhi' not in self._weather_data
+    ):
+      raise ValueError(
+          'Irradiance data not available. Make sure latitude/longitude were'
+          ' provided during initialization.'
+      )
+
+    timestamp = self._ensure_timestamp_tz(timestamp)
```
s2t2: Let's refactor this time check that is repeated across multiple places into a shared method, like 

```
def _validate_timestamp(self, timestamp):
    timestamp = self._ensure_timestamp_tz(timestamp)
    min_time = min(self._weather_data['Time'])
    if timestamp < min_time:
      raise ValueError(
          f'Attempting to get weather data at {timestamp}, before the earliest'
          f' timestamp {min_time}.'
      )
    max_time = max(self._weather_data['Time'])
    if timestamp > max_time:
      raise ValueError(
          f'Attempting to get weather data at {timestamp}, after the latest'
          f' timestamp {max_time}.'
      )
   return timestamp
```

We could fold the time-zone conversion logic (`_ensure_timestamp_tz`) into this function, or leave it separate.

smart_control/simulator/weather_controller.py
```diff
@@ -296,4 +933,165 @@ def get_current_humidity(self, timestamp: pd.Timestamp) -> float:
...
+    Returns:
+      Dictionary with 'ghi', 'dni', 'dhi', 'solar_zenith', and 'solar_azimuth'
+      keys. Irradiance values in W/m2, angles in degrees.
+
+    Raises:
+      ValueError: If latitude/longitude not provided or timestamp out of range
+      or irradiance not calculated.
+
+    Sources:
+      https://pvlib-python.readthedocs.io/en/v0.6.1/_modules/pvlib/forecast.html#ForecastModel.cloud_cover_to_irradiance
```
s2t2: Appreciate this reference link

smart_control/simulator/weather_controller.py
```diff
@@ -160,28 +458,358 @@ def get_outside_air_temp(observation_response):
...
+      # Get clear sky irradiance
+      clearsky = self._location.get_clearsky(datetime_index, model='ineichen')
+
+      # Estimate GHI from cloud cover using linear relationship
+      ghi = clearsky['ghi'].values * (
+          1.0 - 0.8 * (self._weather_data['SkyCoverage'].values / 100.0)
+      )
+
+      # Estimate DNI using DISC model
+      from pvlib.irradiance import disc  # pylint: disable=import-outside-toplevel
```
s2t2: Let's Prefer to have all imports at the top of the file

smart_control/simulator/weather_controller.py
```diff
@@ -160,28 +458,358 @@ def get_outside_air_temp(observation_response):
...
+      return
+
+    # Replace invalid cloud coverage values with NaN
+    self._weather_data.loc[
+        (self._weather_data['SkyCoverage'] < 0)
+        | (self._weather_data['SkyCoverage'] > 100),
+        'SkyCoverage',
+    ] = np.nan
+
+    self._weather_data_ = self._weather_data.copy()
```
s2t2: Can you please explain why we are creating a copy of the dataframe `_weather_data_` and storing it back in `self`? We might prefer to use a separate temporary variable and/or update / mutate the original df.

Due to the variable naming, it also might be hard for human brain to easily distinguish between the two dataframe variable names (`_weather_data_` vs `_weather_data`)

smart_control/simulator/weather_controller.py
```diff
@@ -160,28 +458,358 @@ def get_outside_air_temp(observation_response):
...
+      # Already timezone-aware - convert to controller's timezone
+      return timestamp.tz_convert(self.timezone)
+
+  def _calculate_sky_temperature_column(self):
+    """Pre-calculate sky temperature for all timestamps in weather data.
+
+    Uses Clark & Allen formula with dry bulb and dew point temperatures.
+    """
+    # Stefan-Boltzmann constant
+    sigma = 5.6697e-8  # W/(m^2*K^4)
```
s2t2: I see this constant is used in building radiation utils file as well. Let's define it once in  the simulation/constants.py file, and  import / access from there.

https://github.com/google/sbsim/blob/4f32db25c1ef1b1d25d1abf083050d5e4069e9fb/smart_control/simulator/building_radiation_utils.py#L158

smart_control/simulator/weather_controller.py
```diff
@@ -160,28 +458,358 @@ def get_outside_air_temp(observation_response):
...
+
+    # Get dry bulb temperature in Kelvin (use numpy operations for arrays)
+    if 'TempC' in self._weather_data:
+      temp_k = self._weather_data['TempC'].values + 273.15
```
s2t2: if possible, let's use / apply temp conversion functions from conversion utils (c to k, f to k)

applies for temp and dewpoint below

smart_control/simulator/weather_controller.py
s2t2: If the station_info is now a required parameter for the weather controller, we can pass it as none, and create one using default values if one is not passed:

```
# Args:

station: WeatherStation | None = None

# Init
self.station = station or WeatherStation() # use default values for building SB1

```

smart_control/simulator/weather_controller.py
```diff
@@ -160,28 +458,358 @@ def get_outside_air_temp(observation_response):
...
+    observation_responses: array of observations to extract weather from
+    dewpoint_depression: Difference between dry bulb and dew point temperatures
+      in K. Used if dew point sensor not available. Default is 5.0 K.
+
+  Returns: map from timestamp to sky temperature (K)
+  """
+  # Stefan-Boltzmann constant
+  sigma = 5.6697e-8  # W/(m^2*K^4)
+
+  def get_value(observation_response, measurement_name):
```
s2t2: I see these nested `get_value` methods are repeated. Let's prefer to move them to a single function at the top of the file (outside the class), and invoke from that location.

smart_control/simulator/weather_controller.py
```diff
@@ -160,28 +458,358 @@ def get_outside_air_temp(observation_response):
...
+
+  sky_temps = {}
+  for r in observation_responses:
+    # Get dry bulb temperature
+    temp_k = get_value(r, 'outside_air_temperature_sensor')
+    if temp_k is None:
+      continue
+
+    # Try to get dew point temperature, otherwise estimate from depression
+    dp_k = get_value(r, 'dew_point_temperature_sensor')
```
s2t2: Let's define these sensor names as constants at the top of the file.

smart_control/simulator/weather_controller.py
```diff
@@ -296,4 +933,165 @@ def get_current_humidity(self, timestamp: pd.Timestamp) -> float:
...
+      raise ValueError('SkyCoverage column not found in weather data.')
+
+    times = np.array(self._weather_data.index)
+    target_timestamp = (timestamp - _EPOCH).total_seconds()
+    cloud_cover = np.interp(
+        target_timestamp, times, self._weather_data['SkyCoverage']
+    )
+    return float(cloud_cover)
+
+  def get_current_irradiance(
```
s2t2: If the specific structure of a dictionary containing these specific keys is used in many places, let's consider using a dataclass to define those expected properties. That would also allow us to better document the meaning behind each of these properties.

```py
@dataclasses.dataclass
class Irradiance
  """ Docstring with ...
   Args:
     ...
  """
  ghi: float
  dni: float
  dhi: float
   # etc. 
```

This function can then return an instance of that dataclass.

Later, when we need to reference the values, we can use dot notation (like `irr.ghi`) or convert the dataclass to a dict using `dataclasses.asdict()` and access the values using dictionary notation.

s2t2 (changes requested):
Hi @ecosang thanks for this PR.

I left some comments below. 

One of the major ones is to see if we can use separate test classes to leverage shared set up methods (as applicable), and see if we can use test class inheritance to reduce the amount of test code (if possible).

Also it would be helpful to define some of the expected column names and sensor names as constants.

Also we can see if using some dataclasses would be helpful in specific situations mentioned.

There are some opportunities to refactor shared methods.

Also, let's discuss how to best handle the pvlib dependency, either making it non-optional, or implementing it as optional using the pyproject.toml / poetry mechanism used for this purpose.

smart_control/simulator/local_weather_test_data.csv
ecosang: I put this in "building_radiation_test_data" folder.

smart_control/simulator/weather_controller.py
ecosang: I added it in https://github.com/google/sbsim/pull/165/changes/265540fb09fd9dd03cab83b9381aa2cdd94ef390#diff-0cfd48097b60803cd9e61fff1f3aa765d176d8394319cd0a2eaf449ff32f2dcdR171

smart_control/simulator/weather_controller.py
ecosang: it is done in https://github.com/google/sbsim/commit/265540fb09fd9dd03cab83b9381aa2cdd94ef390#diff-256b9901b7e7079817a790409a91e62dc0c8797bc5bef0b390390652bb82c41aR29

smart_control/simulator/weather_controller.py
ecosang: https://github.com/google/sbsim/commit/265540fb09fd9dd03cab83b9381aa2cdd94ef390#diff-256b9901b7e7079817a790409a91e62dc0c8797bc5bef0b390390652bb82c41aR465

smart_control/simulator/weather_controller.py
ecosang: https://github.com/google/sbsim/commit/265540fb09fd9dd03cab83b9381aa2cdd94ef390#diff-256b9901b7e7079817a790409a91e62dc0c8797bc5bef0b390390652bb82c41aR706

https://github.com/google/sbsim/commit/265540fb09fd9dd03cab83b9381aa2cdd94ef390#diff-256b9901b7e7079817a790409a91e62dc0c8797bc5bef0b390390652bb82c41aR718

https://github.com/google/sbsim/commit/265540fb09fd9dd03cab83b9381aa2cdd94ef390#diff-256b9901b7e7079817a790409a91e62dc0c8797bc5bef0b390390652bb82c41aR742

smart_control/simulator/weather_controller.py
ecosang: https://github.com/google/sbsim/commit/265540fb09fd9dd03cab83b9381aa2cdd94ef390#diff-ba4316e0f4a2ac3741564675aad1c132fbac964047a1f416d1adb4e1c59175acR177

smart_control/simulator/building_radiation_utils.py
ecosang: i updated it..  https://github.com/google/sbsim/pull/165/changes/3bf35f56b2bc648b74ba8be665da868242691e08#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R62

also the lock file.

https://github.com/google/sbsim/pull/165/changes/3bf35f56b2bc648b74ba8be665da868242691e08#diff-f53a023eedfa3fbf2925ec7dc76eecdc954ea94b7e47065393dbad519613dc89R1

Not sure if it is what you suggested.

smart_control/simulator/local_weather_test_data.csv
s2t2: Since the local weather test data is used for replay weather controller tests (without radiation), let's prefer to keep it where it is. Or perhaps we can move both CSV files to a generic "simulator/test_data" directory.

smart_control/simulator/building_radiation_test_data/station.json
```diff
@@ -0,0 +1,12 @@
+{
```
s2t2: Can we use the JSON file from the configs/resources directory, instead of creating a copy of it here? That way we have one source of truth.

https://github.com/google/sbsim/blob/copybara_push/smart_control/configs/resources/sb1/weather_data/station.json

smart_control/simulator/building_radiation_utils.py
```diff
@@ -5,12 +5,15 @@
...
 import math
-from typing import Optional, Tuple
+from typing import Optional, Tuple, TYPE_CHECKING
 
 import numpy as np
 
 from smart_control.simulator import constants
 
+if TYPE_CHECKING:
+  from smart_control.simulator import weather_controller
```
s2t2: Let's do a normal import here.

smart_control/simulator/building_radiation_utils.py
```diff
@@ -967,3 +972,50 @@ def mark_directly_seeing_nodes(
   # Mark the base node with a special value
   floor_plan_copy[base_row, base_col] = blocked_value + marked_value
   return floor_plan_copy
+
+
+def calculate_poa_irradiance(
+    irradiance_components: 'weather_controller.IrradianceComponents',
```
s2t2: Let's use the actual class here for type hint

smart_control/simulator/building_radiation_utils.py
s2t2: Let's just do a normal import of the pvlib at the top of the file. 

It is a small enough dependency, and we are already using scipy, pandas, numpy, etc, so size isn't a big concern.

I know we discussed possibly making it an optional dependency in the poetry / pyproject.toml, but I don't see a strong argument for that, unless there's something I'm missing.

smart_control/simulator/building_radiation_utils.py
```diff
@@ -967,3 +972,50 @@ def mark_directly_seeing_nodes(
...
+  # Import here to avoid circular imports and keep pvlib as optional dependency
+  from pvlib import irradiance as pvlib_irradiance  # pylint: disable=import-outside-toplevel
+
+  poa_irrad = pvlib_irradiance.get_total_irradiance(
+      surface_tilt=surface_tilt,
+      surface_azimuth=surface_azimuth,
+      dni=irradiance_components.dni,
+      ghi=irradiance_components.ghi,
+      dhi=irradiance_components.dhi,
+      solar_zenith=solar_zenith,
```
s2t2: Since solar_zenith and solar azimuth are attributes of the Irradiance Components, would we use those values from the Irradiance Components, instead of passing in separate duplicate params to this function?

smart_control/simulator/constants.py
```diff
@@ -16,40 +16,43 @@
 # pylint: enable=line-too-long
 # fmt: on
 
+
```
s2t2, line 19: /gemini please provide a list of constants added, a list of constants that were rearranged, and a list of constants that were removed (as applicable)

smart_control/simulator/constants.py
```diff
@@ -16,40 +16,43 @@
...
+#################### FILE_INPUT ################################
```
s2t2, line 20: @gemini-code-assist please provide a list of constants added, a list of constants that were rearranged, and a list of constants that were removed (as applicable)

smart_control/simulator/local_weather_test_data.csv
s2t2, line 1: let's move both of the CSV test data files into a "simulator/test_data" directory.

pyproject.toml
```diff
@@ -59,6 +59,14 @@ mkdocs-autorefs = "^1.4.2"
 [tool.poetry.group.notebooks.dependencies]
 jupyter = "^1.1.1"
 
+# pvlib is kept as an optional dependency because it is only needed for solar
+# irradiance calculations (SyntheticWeatherController and
+#ReplayWeatherController with lat/lon).
+# Installing it unconditionally would add a heavy dependency for
+# users who do not use those features. Install with: poetry install --with solar
+[tool.poetry.group.solar.dependencies]
```
s2t2: Sorry for going back and forth on this, but after more investigation, let's prefer to make this a normal dependency (not optional).

smart_control/simulator/simulator.py
```diff
@@ -40,6 +40,8 @@ def __init__(
       iteration_limit: int,
       iteration_warning: int,
       start_timestamp: pd.Timestamp,
+      relative_convergence_threshold: float | None = 1e-6,
```
s2t2, line 43: /gemini how are these new attributes used in this PR?

smart_control/simulator/simulator.py
```diff
@@ -40,6 +40,8 @@ def __init__(
...
+      relative_convergence_streak: int = 20,
```
s2t2, line 44: @gemini-code-assist how are these new attributes used in this PR?

smart_control/simulator/weather_controller.py
```diff
@@ -1,15 +1,34 @@
...
+# for two reasons:
+#
+# 1. Optional dependency: placing the import here would cause an ImportError for
+#    any user who has not installed the "solar" extras group, even if they never
+#    use the irradiance / solar-position features.
+#
+# 2. Circular-import risk: building_radiation_utils imports weather_controller
+#    (guarded by TYPE_CHECKING) to type-hint IrradianceComponents.  If pvlib
+#    were imported at module load time here, and pvlib itself (or a future
+#    transitive dependency) ever imported anything from smart_control, that
```
s2t2: at this time let's assume pvlib won't be importing from smart_control

smart_control/simulator/weather_controller.py
```diff
@@ -1,15 +1,34 @@
...
 import pandas as pd
+# pvlib is an optional dependency (install with: poetry install --with solar).
+# It is imported lazily inside functions rather than at the top of this module
+# for two reasons:
+#
+# 1. Optional dependency: placing the import here would cause an ImportError for
+#    any user who has not installed the "solar" extras group, even if they never
+#    use the irradiance / solar-position features.
+#
+# 2. Circular-import risk: building_radiation_utils imports weather_controller
```
s2t2: if we need to move the IrradianceComponents to avoid a circular import, let's consider moving it to its own file in the simulator directory, perhaps named "building_radiation_models.py". We can have this weather controller and the radiation utils import from there. I think that should avoid a circular import, right?

smart_control/simulator/weather_controller.py
```diff
@@ -125,10 +301,211 @@ def get_air_convection_coefficient(self, timestamp: pd.Timestamp) -> float:
...
+    # Static cloud cover mode
+    if self.cloud_cover is not None:
+      return self.cloud_cover
+
+    # Default: clearsky (no clouds)
+    return 0.0
+
+  def get_current_irradiance(
+      self, timestamp: pd.Timestamp
+  ) -> 'IrradianceComponents':
```
s2t2: Let's use the real class name here as a type hint

smart_control/simulator/weather_controller_test.py
```diff
@@ -122,15 +168,449 @@ def test_get_air_convection_coefficient(self):
...
+        places=4,
+    )
+    self.assertAlmostEqual(
+        irrad.solar_azimuth,
+        solar_position['azimuth'].iloc[0],
+        places=4,
+    )
+
+
+class WeatherControllerIrradianceTest(IrradianceTestBase):
```
s2t2: Nice inheritance of the test classes. 

Since these three Irradiance test classes have so many lines of code, let's move them to their own test files, like "weather_controller_irradiance_test.py" or something like that. This will make them easier to review and maintain.

s2t2 (changes requested):
Hi @ecosang thanks for the updates. I provided another round of comments. 

FYI - in a few comments I am just testing to see if Gemini Code Assist is installed correctly.

smart_control/simulator/weather_controller.py
ecosang: don't understand what it means.

ecosang:
I am kind of lost and can't figure it out. Let's clarify where to put which codes and what to create for a new testing file. Then, I will make a new PR, and I will drop this one.
- Put weather_test file to where?
- Make the sbsim working without pvlib?
- pvlib optional dependency?
- create a new class for the irrdiance test. 
- Split relative_threshold in another PR.
- make irrdiance data and station as new data class.
```
Maybe we can pass an instance of the WeatherStation class I suggested, and use that here.

This would allow us to remove the following params from the initialization:

station_json_path: str | None = None,
timezone: str | None = None,
latitude: float | None = None,
longitude: float | None = None,
In favor of:

self.latitude = self.station.lat
self.longitude = self.station.lon
self.time_zone = self.station.time_zone
s2t2 replied 

s2t2
Collaborator
If the station_info is now a required parameter for the weather controller, we can pass it as none, and create one using default values if one is not passed:

# Args:

station: WeatherStation | None = None

# Init
self.station = station or WeatherStation() # use default values for building SB1

```

- remove _ensure_timestamp_tz
```

def _validate_timestamp(self, timestamp):
    timestamp = self._ensure_timestamp_tz(timestamp)
    min_time = min(self._weather_data['Time'])
    if timestamp < min_time:
      raise ValueError(
          f'Attempting to get weather data at {timestamp}, before the earliest'
          f' timestamp {min_time}.'
      )
    max_time = max(self._weather_data['Time'])
    if timestamp > max_time:
      raise ValueError(
          f'Attempting to get weather data at {timestamp}, after the latest'
          f' timestamp {max_time}.'
      )
   return timestamp
```
- remove self._weather_data_
- define constatns at the top

```

_OUTSIDE_AIR_TEMP_SENSOR: Final[str] = 'outside_air_temperature_sensor'
_DEW_POINT_TEMP_SENSOR: Final[str] = 'dew_point_temperature_sensor'
_CLOUD_COVER_SENSOR: Final[str] = 'cloud_cover_sensor'
_GHI_SENSOR: Final[str] = 'ghi_sensor'
_DNI_SENSOR: Final[str] = 'dni_sensor'
_DHI_SENSOR: Final[str] = 'dhi_sensor'
Let's consider making all the specific important column names into constants, like SKY_TEMP_COL="'TempSkyC'" or SKY_TEMPERATURE_COLUMN_NAME="TempSkyC"

```
- Put irradiance stuff in weather_controller by default? or all optional? then how to deal with pvlib?


Anything I missed?

smart_control/simulator/building_radiation_test_data/station.json
ecosang: The main reason is station depends on the weather file. So, csv file should be with the station.json.

smart_control/simulator/weather_controller.py
ecosang: _weather_data is the one that is used in calculation. while we need to _weather_data_ for calculaing clearsky.. I think I don't need to have _weather_data_ in self.

s2t2:
"simulation/solar_radiation.py"


```py
import dataclasses

import pvlib
from pvlib.irradiance import campbell_norman  
from pvlib.irradiance import get_extra_radiation  
from pvlib import location

IRRADIANCE_METHODS = ('clearsky', 'linear', or 'campbell_norman')

GHI_SENSOR: Final[str] = 'ghi_sensor'
DNI_SENSOR: Final[str] = 'dni_sensor'
DHI_SENSOR: Final[str] = 'dhi_sensor'


@dataclasses.dataclass
class BuildingInfo:
  """Information about the building under control."""
  floor_plan_filepath: str = "...." # path to the building's floor plan npy file
  floor_plan_orientation: float = 0  # angle of floor plan orientation (0/360 = North)
  lat: float = 37.4263  # latitude of the building
  lon: float = -122.0349 # longitude of the building  
  time_zone: str = "US/Pacific"
  altitude: float | None = None

  def __post_init__:
     self._validate_floorplan_orientation

  def validate_floorplan_orientation(self):
     if self.floorplan_orientation < 0 or self.floorplan_orientation > 360:
        raise ValueError(f"Expecting floorplan orientation to be between 0-360, but got {self.floorplan_orientation}.")
    

# notes from pvlib.Location
# altitude ([float](https://docs.python.org/3/library/functions.html#float), optional) – Altitude from sea level in meters. If not specified, the altitude will be fetched from


@dataclasses.dataclass
class IrradianceComponents:
  """Irradiance components and solar position at a given timestamp.

  Args:
    ghi: Global Horizontal Irradiance in W/m^2. Total solar radiation
      received on a horizontal surface.
    dni: Direct Normal Irradiance in W/m^2. Solar radiation received
      perpendicular to the sun's rays.
    dhi: Diffuse Horizontal Irradiance in W/m^2. Solar radiation received
      on a horizontal surface from the sky (excluding direct beam).
    solar_zenith: Solar zenith angle in degrees (angle from vertical).
      0 = sun directly overhead, 90 = sun at horizon.
    solar_azimuth: Solar azimuth angle in degrees (compass direction of sun).
      0/360 = North, 90 = East, 180 = South, 270 = West.
  """
   timestamp: pd.Timestamp
  ghi: float
  dni: float
  dhi: float
  solar_zenith: float # optional?
  solar_azimuth: float # optional


class SolarRadiation():
  """
   
   Calculates solar radiation, depending on the cloud cover, and building location (lat/lon and time zone).

  """
  def __init__(self, 
    building_info: BuildingInfo | None = None,
    weather_controller: BaseWeatherController,
    outdoor_air_temp: float, # check name to be consistent with weather controller. Docs: in Kelvin!
    dewpoint_depression: float = 5.0,
    cloud_cover: float | None = None,
    cloud_cover_low: float | None = None,
    cloud_cover_high: float | None = None,
    irradiance_method: str = 'clearsky',

):
    self.building_info = building_info or BuildingInfo()
    self.time_zone = self.building_info.time_zone
    self.lat = self.building_info.lat
    self.lon = self.building_info.lon

    self.weather_controller = weather_controller

    self.cloud_cover ...
    # ...
    self.irradience_method ...

def get_replay_irradiance(
    observation_responses: Sequence[
        smart_control_building_pb2.ObservationResponse
    ],
) -> Sequence[IrradianceComponents]:
  irradiances = []
  for r in observation_responses:
    ghi = _get_observation_value(r, _GHI_SENSOR, default=0.0)
    dni = _get_observation_value(r, _DNI_SENSOR, default=0.0)
    dhi = _get_observation_value(r, _DHI_SENSOR, default=0.0)

    timestamp = utils.proto_to_pandas_timestamp(r.timestamp)
    irradiances.append(
      IrradianceComponents(
         timestamp=timestamp,
          ghi= ghi,
          dni=dni,
          dhi=dhi,
          solar_zenith = ? # TBD
          solar_azimuth = ? # TBD
    ))

  return irradiances

def get_current_sky_temperature(self, timestamp: pd.Timestamp) -> float:
    """Returns sky temperature in K using Clark & Allen formula.

    Args:
      timestamp: Pandas timestamp to get sky temperature for. If naive
        (no timezone), will be localized to the controller's timezone.

    Returns:
      Sky temperature in K.
    """
    timestamp = self._ensure_timestamp_tz(timestamp)

    sigma = constants.STEFAN_BOLTZMANN_CONSTANT

    # Get dry bulb temperature (timestamp already localized)
    temp_k = self.weather_controller.get_current_temp(timestamp)

    # Estimate dew point temperature
    dp_k = temp_k - self.dewpoint_depression

    # Calculate sky emissivity (Clark & Allen)
    epsilon_sky = 0.787 + 0.764 * np.log(dp_k / 273.0)

    # Calculate horizontal infrared radiation
    ir_h = epsilon_sky * sigma * (temp_k**4)

    # Calculate sky temperature
    temp_sky_k = (ir_h / sigma) ** 0.25

    return temp_sky_k

def _get_observation_value(
    observation_response: smart_control_building_pb2.ObservationResponse,
    measurement_name: str,
    default=None,
):
  """Returns the continuous value for a named measurement in an observation.

  Searches the single_observation_responses of the given observation_response
  for an entry whose measurement_name matches the requested name.

  Args:
    observation_response: A single ObservationResponse proto.
    measurement_name: The sensor/measurement name to look up.
    default: Value to return when the measurement is not found. Defaults to
      None.

  Returns:
    The continuous_value of the matching observation, or ``default`` if no
    matching measurement is found.
  """
  for r in observation_response.single_observation_responses:
    if r.single_observation_request.measurement_name == measurement_name:
      return r.continuous_value
  return default




```

s2t2:
For tests


"simulation/solar_radiation_test.py"


```py

class SolarRadiationTest(absltest.TestCase):
    # use the normal weather controller

    def setUp():
      self.weather_controller = WeatherController(...)
      self.solar_radiation = SolarRadiation(weather_controller=self.weather_controller, ...)
 
   # all the tests (should apply for whatever weather controller we're using)


class ReplaySolarRadiationTest(SolarRadiationTest)
   # use the replay weather controller

   def setUp():
      self.weather_controller = ReplayWeatherController(...)
      self.solar_radiation = SolarRadiation(weather_controller=self.weather_controller, ...)
 
   # since we are inheriting, we might not need any / many new test methods here
   # all the ones in the parent class will be run

   # if one of the parent tests doesn't apply to this replay child class
   # we can override that test method

```


If we need to add new test data files, let's do it in a new "simulation/solar_radiation_test_data" directory.

ecosang:
@s2t2 sorry for the delay. I have updated it.

smart_control/simulator/weather_controller_test.py
```diff
@@ -127,7 +127,9 @@ class ReplayWeatherControllerTest(parameterized.TestCase):
   def setUp(self):
     super().setUp()
     data_path = os.path.join(
-        os.path.dirname(__file__), 'local_weather_test_data.csv'
+        os.path.dirname(__file__),
+        'solar_radiation_test_data',
```
s2t2: Let's keep this CSV data at its original location. In the solar radiation tests we can read it from its original location.

smart_control/simulator/solar_radiation.py
```diff
@@ -0,0 +1,649 @@
+"""Solar Radiation Calculations for Building Simulation.
+
+For computing irradiance components, solar position, and sky temperature.
+"""
+
+from __future__ import annotations
```
s2t2: Is this necessary / used?

smart_control/simulator/solar_radiation.py
```diff
@@ -0,0 +1,649 @@
...
+
+import numpy as np
+import pandas as pd
+from pvlib import irradiance as pvlib_irradiance
+from pvlib import location as pvlib_location
+
+from smart_control.simulator import constants
+from smart_control.utils import conversion_utils as utils
+
+if TYPE_CHECKING:
```
s2t2: Is TYPE_CHECKING necessary? Do we have a circular dependency somewhere? If so we should prefer to resolve the underlying issue.

I would think the dependency we want is this solar radiation depending on the weather controller (which I believe is how it is currently set up). Let me know if I'm missing something.

smart_control/simulator/solar_radiation.py
```diff
@@ -0,0 +1,649 @@
...
+  lat: float = 37.4263
+  lon: float = -122.0349
+  time_zone: str = 'US/Pacific'
+  altitude: float | None = None
+
+  def __post_init__(self):
+    self._validate_floor_plan_orientation()
+
+  def _validate_floor_plan_orientation(self) -> None:
+    """Raise if ``floor_plan_orientation`` is outside [0, 360]."""
```
s2t2: FYI - I think we only need one set of backticks for code to show up as markdown in the docs.

smart_control/simulator/solar_radiation.py
```diff
@@ -0,0 +1,649 @@
...
+# IrradianceComponents
+# ---------------------------------------------------------------------------
+@dataclasses.dataclass
+class IrradianceComponents:
+  """Irradiance components and solar position at a given timestamp.
+
+  Attributes:
+    timestamp: Pandas timestamp for the measurement.  May be *None* when
+      the timestamp is not available (e.g. when constructed from a weather
+      controller that does not track it).
```
s2t2, line 93: Awesome documentation throughout this file. Thank you!

smart_control/simulator/solar_radiation.py
```diff
@@ -0,0 +1,649 @@
...
+  irradiance: IrradianceComponents
+
+
+# ---------------------------------------------------------------------------
+# SolarRadiation calculator
+# ---------------------------------------------------------------------------
+class SolarRadiation:
+  """Calculates solar radiation for a building location.
+
+  Combines a :class:`simulator.weather_controller.BaseWeatherController`
```
s2t2, line 144: @gemini-code-assist Just curious - does this `:class:` annotation help with hyperlinking?

smart_control/simulator/solar_radiation.py
```diff
@@ -0,0 +1,649 @@
...
+    self.lon = self.building_info.lon
+    self.weather_controller = weather_controller
+    self.dewpoint_depression = dewpoint_depression
+    self.cloud_cover = cloud_cover
+    self.cloud_cover_low = cloud_cover_low
+    self.cloud_cover_high = cloud_cover_high
+    self.irradiance_method = irradiance_method
+
+    # --- validation -------------------------------------------------------
+    if self.irradiance_method not in IRRADIANCE_METHODS:
```
s2t2, line 192: Let's consider putting these validations in their own protected methods, like:

  + `self._validate_irradience_method()`
  + `self._validate_cloud_cover()`

also maybe a protected constructor method: 

  + `self._location = self._get_pvlib_location()`

smart_control/simulator/solar_radiation_test.py
```diff
@@ -0,0 +1,1229 @@
...
+from smart_control.simulator import building_radiation_utils
+from smart_control.simulator import constants as sim_constants
+from smart_control.simulator import solar_radiation
+from smart_control.simulator import weather_controller
+from smart_control.utils import conversion_utils as utils
+
+# Paths to shared weather data files.
+_LOCAL_WEATHER_TEST_DATA_PATH = os.path.join(
+    os.path.dirname(__file__),
+    'solar_radiation_test_data',
```
s2t2: Let's prefer to keep this test data where it was

smart_control/simulator/solar_radiation_test.py
```diff
@@ -0,0 +1,1229 @@
...
+    )
+
+
+# ---------------------------------------------------------------------------
+# Replay helper function tests
+# ---------------------------------------------------------------------------
+class _ObservationHelper:
+  """Mixin providing _make_observation_response for test classes."""
+
+  def _make_observation_response(
```
s2t2: Because this method does not use `self`, let's prefer to use a module level function (defined on the top level, not inside any class), and we can just call that function from all the tests that need it (instead of inheriting those tests from `_ObservationHelper`).

smart_control/simulator/solar_radiation_test.py
```diff
@@ -0,0 +1,1229 @@
...
+class ReplaySolarRadiationTest(SolarRadiationTest):
+  """Re-runs all SolarRadiationTest tests with ReplayWeatherController."""
+
+  def setUp(self):
+    # Skip SolarRadiationTest.setUp; call IrradianceTestBase.setUp directly.
+    super(SolarRadiationTest, self).setUp()  # pylint: disable=bad-super-call
+
+    # Clear any gin bindings from previous tests that might override
+    # ReplayWeatherController's local_weather_path default.
+    gin.clear_config()
```
s2t2: Are we using gin to set values during testing? 

I'm not aware of gin being used to setup tests, but if I am missing something, let me know.

If we aren't using gin to setup tests, let's remove this.

s2t2 (changes requested):
@ecosang this PR is looking great overall. Awesome tests and documentation.

I had a few minor comments, and then LGTM once they are addressed.

Thank you!

smart_control/simulator/solar_radiation.py
ecosang: While migrating my previous PR, it was imported unnecessarily. I reviewed this file again.

smart_control/simulator/solar_radiation_test.py
ecosang: @s2t2 I think I've fixed your comments.

docs/api/simulator/solar_radiation.md
```diff
@@ -0,0 +1,3 @@
+# Solar Radiation
```
s2t2, line 1: @gemini-code-assist do we need to update another file in the docs site to add this new docs page to the navigation? 

we are using mkdocstrings

smart_control/simulator/solar_radiation_test_data/723170TYA.CSV
s2t2, line 1: @ecosang the first few lines of this CSV file are pasted below. Is the first line a formatting issue? Or is it ok to leave as is?

```csv
723170,"GREENSBORO PIEDMONT TRIAD INT",NC,-5.0,36.100,-79.950,273
Date (MM/DD/YYYY),Time (HH:MM),ETR (W/m^2),ETRN (W/m^2),GHI (W/m^2),GHI source,GHI uncert (%),DNI (W/m^2),DNI source,DNI uncert (%),DHI (W/m^2),DHI source,DHI uncert (%),GH illum (lx),GH illum source,Global illum uncert (%),DN illum (lx),DN illum source,DN illum uncert (%),DH illum (lx),DH illum source,DH illum uncert (%),Zenith lum (cd/m^2),Zenith lum source,Zenith lum uncert (%),TotCld (tenths),TotCld source,TotCld uncert (code),OpqCld (tenths),OpqCld source,OpqCld uncert (code),Dry-bulb (C),Dry-bulb source,Dry-bulb uncert (code),Dew-point (C),Dew-point source,Dew-point uncert (code),RHum (%),RHum source,RHum uncert (code),Pressure (mbar),Pressure source,Pressure uncert (code),Wdir (degrees),Wdir source,Wdir uncert (code),Wspd (m/s),Wspd source,Wspd uncert (code),Hvis (m),Hvis source,Hvis uncert (code),CeilHgt (m),CeilHgt source,CeilHgt uncert (code),Pwat (cm),Pwat source,Pwat uncert (code),AOD (unitless),AOD source,AOD uncert (code),Alb (unitless),Alb source,Alb uncert (code),Lprecip depth (mm),Lprecip quantity (hr),Lprecip source,Lprecip uncert (code),PresWth (METAR code),PresWth source,PresWth uncert (code)
01/01/1988,01:00,0,0,0,1,0,0,1,0,0,1,0,0,1,0,0,1,0,0,1,0,0,1,0,10,A,7,10,A,7,10.0,A,7,6.1,A,7,77,A,7,993,A,7,200,A,7,6.2,A,7,16100,B,7,1370,A,7,1.5,E,8,0.000,?,0,0.00,?,0,0,1,D,9,00,C,8
01/01/1988,02:00,0,0,0,1,0,0,1,0,0,1,0,0,1,0,0,1,0,0,1,0,0,1,0,10,A,7,10,A,7,10.0,A,7,6.7,A,7,80,A,7,993,A,7,230,A,7,5.2,A,7,16100,B,7,1370,A,7,1.6,E,8,0.000,?,0,0.00,?,0,0,1,D,9,00,C,8
01/01/1988,03:00,0,0,0,1,0,0,1,0,0,1,0,0,1,0,0,1,0,0,1,0,0,1,0,10,A,7,10,A,7,10.0,A,7,7.2,A,7,83,A,7,993,A,7,220,A,7,5.7,A,7,16100,B,7,1370,A,7,1.6,E,8,0.000,?,0,0.00,?,0,0,1,D,9,00,C,8
01/01/1988,04:00,0,0,0,1,0,0,1,0,0,1,0,0,1,0,0,1,0,0,1,0,0,1,0,10,A,7,10,A,7,10.0,A,7,7.2,A,7,83,A,7,992,A,7,210,A,7,5.7,A,7,16100,B,7,1370,A,7,1.6,E,8,0.000,?,0,0.00,?,0,0,1,D,9,00,C,8

```

smart_control/simulator/solar_radiation.py
```diff
@@ -0,0 +1,659 @@
...
+_DEW_POINT_TEMP_SENSOR: Final[str] = 'dew_point_temperature_sensor'
+_CLOUD_COVER_SENSOR: Final[str] = 'cloud_cover_sensor'
+
+
+# ---------------------------------------------------------------------------
+# BuildingInfo
+# ---------------------------------------------------------------------------
+@dataclasses.dataclass
+class BuildingInfo:
+  """Information about the building under control.
```
s2t2, line 47: @gemini-code-assist Is this Docstring structured optimally? Should we add a raises section? Or is that only reserved for function / method docstrings?

smart_control/simulator/solar_radiation.py
```diff
@@ -0,0 +1,659 @@
...
+    observation_response: A single `ObservationResponse` proto.
+    measurement_name: The sensor / measurement name to look up.
+    default: Value to return when the measurement is not found.
+
+  Returns:
+    The `continuous_value` of the matching observation, or *default* if no
+    matching measurement is found.
+  """
+  for r in observation_response.single_observation_responses:
+    if r.single_observation_request.measurement_name == measurement_name:
```
s2t2, line 486: FYI I believe in some internal updates, where there are two air handlers, there can be two different outdoor air temp sensors, and in some cases we have changed this equality check to an inclusion check:

`if measurement_name in r.single_observation_request.measurement_name`

However I think this is OK to keep for now until we have synced the internal codebase out.

s2t2 (approved):
@ecosang thank you for all the work you have put into updating this PR (going back to the original large PR where this all started). 🙏 

LGTM!

s2t2:
@ecosang Let me know about this CSV file comment / question, and then when I hear back from you we can merge
https://github.com/google/sbsim/pull/165#discussion_r3169846551

smart_control/simulator/solar_radiation.py
ecosang: interesting.  Typically, we use single outdoor air temp for a single building. but that means we need to define multiple weather controllers for multiple ahus.

smart_control/simulator/solar_radiation_test_data/723170TYA.CSV
ecosang: this is TMY3 format, that is used for building energy simulation in general. I don't like this format, but there is a function to read this kind of file using pvlib. I have added the details here.  https://github.com/google/sbsim/pull/165/changes#diff-2f9cfed8705a27c0e49d54e817370321ee8e8ade4750e0639a676e8d6f2c7bc6R822

s2t2: OK, thank you!

PR #127, opened 2025-10-14
Yukta/rl module finish

yuktakul04:
Depends on #98
## Summary
Fix RL training + replay buffers for portability, reproducibility, and testing. **Builds on #98.**

## Changes
- **Replay buffer integration** (`train.py`): switched collection to `py_driver.PyDriver`; buffer grows correctly; pass `time_step` to `collect_driver.run()`.
- **Reproducibility**: new `seed.py` utility; seed `random`, `numpy`, `tensorflow` (seed=42).
- **Robust paths**: create dirs proactively; avoid `FileExistsError` on re-runs; stable results layout.
- **Dataset creation**: use `replay_manager.get_dataset()` (uniform fallback-friendly).
- **Replay buffer portability** (`replay_buffer.py`):
  - Make **dm-reverb optional** (macOS/py3.11 lacks wheels).
  - **Fallback to `TFUniformReplayBuffer`** when Reverb isn’t present.
  - Wrap TFUniform observer to **add a batch dimension** (fixes `InvalidArgumentError`).
- **Tests**:
  - `tests/test_replay_buffer_manager.py`  (now PASS)
  - `tests/test_train_script_smoke.py` (optional, guarded via `RUN_SLOW_TESTS=1`)

## Verification
- Unit test: `pytest tests/test_replay_buffer_manager.py -v` → **1 passed** (TF/Gym deprecations only).
- Replay buffer size increments as expected.
- End-to-end run saves checkpoints/results to `data/experiment_results/<exp_name>/`.

## Notes
- On macOS/Py3.11, `dm-reverb` isn’t installable; the code gracefully uses TFUniform.
- Maintainers can enable Reverb-backed buffers on Linux by installing `dm-reverb`.

PR #166, opened 2026-04-26
Finish PR #98: replay fallback checkpoint compatibility + RL test updates

amcberkes:
- Completes PR#98
- Incorporates replay fallback direction from #127
- Replaces Reverb-only checkpoint calls with ReplayBufferManager.save_checkpoint()
- Adds TFUniform fallback checkpoint save/restore support
- Updates RL tests for fallback compatibility


amcberkes:
I got the RL training running end-to-end on my Mac without dm-reverb.

I switched replay handling to a TFUniform fallback path and added metadata persistence for replay capacity (replay_buffer_metadata.json), which fixed the restore mismatch I was hitting during train startup.

s2t2:
@amcberkes awesome!

Looks like there are two file conflicts that need resolution. Let me know once they are resolved and I will provide a review.

Issue #102, opened 2025-06-20
Fix installation of `dm-reverb` on Mac


s2t2:
We are [noticing](https://github.com/google/sbsim/pull/93) some issues trying to install the `dm-reverb` package on Mac. 

Looks like the `dm-reverb` package may only be meant for Linux. Alternatively we have tried `dm-reverb-macos` and `dm-reverb-macos-nightly`, but ran into issues. 

  + https://github.com/google-deepmind/reverb/issues/11
  + https://github.com/google-deepmind/reverb/pull/128
  + https://github.com/google-deepmind/reverb/pull/119

It would be great if we could fix these issues to enable users and contributors to run the full set of functionality on Mac OS. If we can get things verified to work on both styles of Mac (intel-based and arm-based), that would be ideal.

s2t2:
Here are some details about the installation issues.

In each of these cases we are seeing `ModuleNotFoundError: No module named 'reverb'` 

## dm-reverb-macos (0.11.0.dev0)

Configuration:

```
# pyproject.toml ...

dm-reverb-macos = { version = "0.11.0.dev0", markers = "sys_platform == 'darwin'" }
```

**Results on macos-latest (arm-based):**

https://github.com/google/sbsim/actions/runs/15783794271/job/44495589031

![Image](https://github.com/user-attachments/assets/46043075-4075-4bc0-80da-0926194e94d7)


**Results on macos-13 (intel-based):**

https://github.com/google/sbsim/actions/runs/15783794271/job/44496055041

![Image](https://github.com/user-attachments/assets/f7fbafe1-777f-46f5-afa3-b11230a5d5e0)









<hr>

## dm-reverb-macos-nightly (>=0.11.20241207)

Configuration:

```
# pyproject.toml ...

dm-reverb-macos-nightly = { version = ">=0.11.20241207", markers = "sys_platform == 'darwin'" }
```

**Results on macos-latest (arm-based):**

https://github.com/google/sbsim/actions/runs/15637935762/job/44057877216

![Image](https://github.com/user-attachments/assets/00a50285-9ea1-4617-ab88-7c705da18f16)


**Results on macos-13 (intel-based):**

https://github.com/google/sbsim/actions/runs/15637935762/job/44490876291

![Image](https://github.com/user-attachments/assets/4817228c-dd46-444b-aeb1-106298b1a32b)

LucasBrazCappelo:
[...]

PR #92, opened 2025-06-05
Update Protos

s2t2:
Generate protos using libprotoc 3.21.12.

To minimize differences in setups, we are pinning a specific version of protoc. The latest version of protoc installed on Linux machines is 3.21.12 (the chosen version). The latest version installed on Mac via homebrew is 3.29 (however this is ahead of the linux installation). It is unclear which version was used to generate the current protos.

PR #173, opened 2026-08-05
feat: Optimize Dockerfile and update setup documentation

### LINKED ISSUE #80: Optimize Dockerfile

ajkhabbazi:
Fixes #80

Supersedes #111 (carries @igopalakrishna's original work forward).

Two changes on top:
- Merged `copybara_push` to bring the branch up to date (it was 10 commits behind). Only `poetry.lock` conflicted; regenerated it from the merged `pyproject.toml` rather than resolving by hand.
- Reverted the three auto-generated `*_pb2.py` files to upstream, per @s2t2's review. They are regenerated at build time by the Dockerfile, so nothing is lost.

Verified: image builds clean, `pytest` passes in the container (830 passed, 37 skipped), and all three `*_pb2.py` files are generated inside the image and import correctly.

- [x] Tests pass


s2t2:
Hi @ajkhabbazi thanks for this update.

We are seeing some failing workflow checks related to the pb2 files - if we are updating the version of the Protos, we should regenerate the pb2 files. I see you excluded them, but it seems we may need to include them. If we include the new proto versions and all tests are passing, we can be confident in the newly generated files.

FYI - I had undergone a similar effort in the past to update the pb2 files and the Linux dependency (not the Python dependency), but never checked in the updated Protos. We probably want to also update the version of the underlying `protoc` linux library (see setup docs), and use it to regenerate the Protos.

See #92 

Also, if we are able to upgrade to Python version 3.14, that would match Google internal Python version, which just got bumped. Please let me know if you are able to take on the Python version update as well. Either as part of this PR or separate follow-on.

ajkhabbazi:
@s2t2 Thanks -- regenerated and pushed.

All three *_pb2.py files are back in, generated with libprotoc 3.21.12. They come out byte-identical to the ones in #92, so that confirms your earlier regeneration was right. Worth closing #92 as superseded once this lands -- it does have the protos committed, it just never merged.

On pinning protoc, I think we're already there. CI installs 3.21.12 on both platforms today and docs/setup/mac.md documents it from your #93, so the checked-in files now just match what everyone has.

One thing I hit along the way. The Dockerfile was regenerating with grpc_tools.protoc, which bundles libprotoc 25.1, so the container produced different gencode than everywhere else. Switched it to the apt protoc already in the image, which needed libprotobuf-dev for the well-known types. grpcio-tools is dropped. Container output now matches the committed files exactly.

The workflow is awaiting approval whenever you get a chance.

On Python 3.14, that one isn't possible in this PR. tensorflow is pinned at 2.15.1, tf-agents to that, and dm-reverb has no wheels past 3.11 (#102), so it needs a TensorFlow upgrade first. Happy to look into it separately once this merges.

s2t2:
@ajkhabbazi thanks for taking this on! 💪 

LGTM though I'm not very knowledgeable about docker.

Question - is there a GitHub workflow or other test we can use to ensure the docker builds successfully? I am creating a separate issue #178 for this.

PR #176, opened 2026-08-10
Preserve MkDocs syntax during Markdown formatting

### LINKED ISSUE #103: Re-evaluate Usage of `mdformat` (Enable and Disable via Comments)
Currently `mdformat` will automatically re-format markdown used by `mkdocstrings`, as well as `mkdocs` [admonitions](https://squidfunk.github.io/mkdocs-material/reference/admonitions/), in a way that breaks their functionality.

Example desired `mkdocstrings` markdown:

```
::: smart_control.dataset.dataset.BuildingDataset
    options:
      members_order: source
      group_by_category: false
```

Example desired `mkdocs` admonition markdown:

```
!!! info "Important: Installing Pre-commit Hooks"

    To enable the pre-commit hooks, you must perform a one-time setup by running
    `pre-commit install`. This will create or update ".git/hooks/pre-commit".
```

Currently there is no way of disabling `mdformat` on a line-by-line basis. So we are currently a) ignoring the `mkdocstrings` related files from being formatted by `mdformat`, and b) not using admonitions.

See: 
  + https://github.com/hukkin/mdformat/issues/436
  + https://github.com/hukkin/mdformat/issues/53

However if `mdformat` implemented comments to enable and disable specific lines, that would solve both of these issues.

Without the ability to enable and disable specific lines from being formatted, if we want to solve these issues we will have to discontinue use of `mdformat` in favor of another tool.

We might want to investigate a different tool for markdown formatting (revisiting https://github.com/google/sbsim/issues/58).

hugosmoreira:
Fixes #103

## Summary

- Upgrade to `mdformat` 1.0.0 and add `mdformat-mkdocs` 5.3.0 in Poetry and pre-commit.
- Format and check the complete `docs` tree, including the previously excluded `docs/api` files.
- Enable MkDocs admonitions and use the issue's example in the contributor guide.
- Update the formatter documentation and regenerate the lockfile with the repository's documented Poetry 2.1.2.

Plain `mdformat` changes both MkDocs admonitions and configured `mkdocstrings` directives in ways that break their structure. With `mdformat-mkdocs`, the same regression fixture is byte-for-byte unchanged.

## Validation

- [x] `poetry check --lock`
- [x] `mdformat README.md docs --check`
- [x] Exact pinned `pre-commit run mdformat --all-files`
- [x] `mkdocs build --clean` (existing Griffe/link warnings remain)
- [x] Generated HTML contains the expected rendered admonition and API documentation
- [x] `git diff --check`
- [ ] Full Python test suite (the locked `tensorflow-io-gcs-filesystem==0.37.1` has no compatible Windows wheel; Linux CI is the authoritative run)


s2t2:
Hi @hugosmoreira thanks for working on this. Let me know when its ready for review. 

Also if there are any screenshots you can add of the locally-built docs site, that will be helpful for acceptance testing.

hugosmoreira:
Thanks! This is ready for review. I rebuilt the documentation site from the current PR head and confirmed that the MkDocs admonition and generated API reference render correctly. I’ve attached screenshots of both views.

sbsim-pr-176-admonition
sbsim-pr-176-api-reference

.github/workflows/build.yaml
```diff
@@ -32,10 +35,10 @@ jobs:
 
     steps:
     - name: Checkout code
-      uses: actions/checkout@v4
+      uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4
```
s2t2, line 38: is there a reason we are pinning to very specific versions rather than the general v4? was there an issue with v4?

.pre-commit-config.yaml
```diff
@@ -33,8 +33,9 @@ repos:
       exclude: '^smart_control/proto/'
 
 # See: https://mdformat.readthedocs.io/en/stable/users/installation_and_usage.html#usage-as-a-pre-commit-hook
-- repo: https://github.com/hukkin/mdformat
-  rev: 0.7.22
+- repo: https://github.com/executablebooks/mdformat
```
s2t2: I noticed that executablebooks owned repo redirects / resolves to the hukkin owned repo. 

Is there a reason this needs to change or would we want to change back to the original hukkin repo?

s2t2 (changes requested):
@hugosmoreira thanks for taking this on!

This PR looks generally good to me. 

Could you please reply to the two minor comments I left? 

Then we can merge.

.github/workflows/build.yaml
hugosmoreira: Good question. There was no functional issue with v4. The initial commit used actions/checkout@v4 and actions/setup-python@v4, but the repository security scan reported zizmor/unpinned-uses for both and stated that a full commit hash was required by its blanket policy. The pinned commits are still v4; using full SHAs makes the references immutable and allowed the mandatory scan to pass. I therefore kept the SHA pins.

.pre-commit-config.yaml
hugosmoreira: Agreed. GitHub resolves executablebooks/mdformat to the canonical hukkin/mdformat repository, and both URLs resolve tag 1.0.0 to the same commit. The URL change was unnecessary, so I restored https://github.com/hukkin/mdformat in signed commit 27dc976 while keeping rev 1.0.0 and mdformat-mkdocs==5.3.0. The exact pre-commit hook passed on all Markdown files and the MkDocs build completed successfully.

s2t2 (approved):

PR #179, opened 2026-08-20
Add Docker image build to CI

### LINKED ISSUE #178: Docker Build in CI Workflow
## Building the Docker Image

It would be nice to implement an automated workflow (likely using GitHub Actions) to ensure our Docker file builds successfully. The goal is to ensure that the Dockerfile leads to a successful build. Another goal would be to use this workflow when making changes to the Dockerfile, to ensure things don't break.

Need to search for, and implement a Docker related GitHub Actions workflow.

And then add it to our CI pipeline.

---

## Deploying Docker Image

As a follow-on, it might be nice to deploy the docker image somewhere (like Docker Hub, GitHub registry, etc) so people can use it more widely.

There might be a separate GitHub Actions workflow for this as well. The action would be triggered by releases (instead of pushes / merges).

danielgaskins:
Fixes #178

Adds a dedicated GitHub Actions workflow that builds the Docker image on pushes and pull requests targeting `master`, `main`, or `copybara_push`.

The workflow uses Docker Buildx, does not publish the image, and caches dependency layers between builds. Cache export failures remain nonfatal so they cannot hide an image build failure.

## Testing

- `actionlint .github/workflows/docker-build.yml`
- `zizmor --format plain .github/workflows/docker-build.yml`

A full image build requires Docker and will run in the new CI job.

- [x] Relevant local checks pass
- [x] No documentation changes are needed

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -1,0 +1,38 @@
+name: Docker CI Build
+
+on:
+  push:
+    branches:
+      - master
+      - main
+      - copybara_push
+  pull_request:
+    branches:
+      - master
+      - main
+      - copybara_push
+
+permissions:
+  contents: read
+
+jobs:
+  build:
+    name: Build Docker image
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v4
+
+      - name: Build Docker image
+        uses: docker/build-push-action@v7
+        with:
+          context: .
+          platforms: linux/amd64
+          push: false
+          tags: sbsim:ci
+          cache-from: type=gha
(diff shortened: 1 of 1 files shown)


s2t2:
Hi @danielgaskins this looks good but I'm just seeing this now and I have just merged #180. Sorry about that.

If there are any aspects of this PR that aren't covered in that other PR, let me know and we can do a follow-up. Maybe you could do the deployment of the docker image to docker hub aspect of #178. We will want to trigger than on separate release (not on PR / merge).

Thanks for your contribution!

danielgaskins:
Thanks, I’m happy to handle this one. Which Docker Hub repository should receive the image, and what secret names should the workflow use for the username and token?

I’d trigger publishing only when a GitHub release is published, tag the image with the release tag, add `latest` only for non-prerelease versions, pin the actions by commit SHA, and leave pull request builds in the existing workflow.

PR #181, opened 2026-08-21
Configure Pyrefly tooling

yash2006kr:
Addresses google/sbsim#174.

## Summary
- add the Google OSS Pyrefly configuration for `smart_control`
- add Pyrefly to the Poetry dev dependency group and regenerate `poetry.lock`
- run Pyrefly from GitHub Actions and pre-commit hooks in report-only mode
- document the new type-checking workflow for contributors

## Notes
Pyrefly currently reports existing findings in the repository, so this PR wires the check in as non-blocking for now. This matches the discussion in #174/#175 about landing the tooling before the upcoming internal sync-out resolves the current findings. The TODO comments mark the CI and pre-commit places where `|| true` / report-only behavior can be removed later.

## Validation
- `poetry lock`
- `poetry check` (passes with existing Poetry deprecation warnings)
- `pyrefly check --config pyrefly.toml` (command/config verified; reports existing findings)

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -77,6 +77,13 @@ jobs:
         source .venv/bin/activate
         pylint --rcfile=.pylintrc --ignore=proto smart_control
 
+    - name: Run the Python type checker
+      run: |
+        source .venv/bin/activate
+        # TODO(#174): Remove "|| true" after the upcoming internal code sync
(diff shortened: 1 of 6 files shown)


s2t2:
Hi @yash2006kr Looks good to me. Appreciate the integrations with the CI build and pre-commit hooks, and the thoughtful way you suggest circling back to make this a stricter check after the upcoming code sync-out.

PR #182, opened 2026-08-21
Move seconds_to_rads to a standalone function

### LINKED ISSUE #117: Convert `WeatherController.seconds_to_rads` to a Static Method
The `seconds_to_rad` method in the `WeatherController` doesn't use `self`, so let's convert it to a static method.

https://github.com/google/sbsim/blob/67c5bb018e54295f59ab2060962bbbeeb5db8412/smart_control/simulator/weather_controller.py#L69-L77

Mustaphayinka:
Fixes #117

Moves secondstorads out of WeatherController and into a standalone, module-level function, following your follow-up advice to use a standalone function instead of a static method. 

Changes: The function now sits above BaseWeatherController, and the nonexistent self parameter has been removed. WeatherController.getcurrenttemp is called directly at its call site rather than through self. In weathercontrollertest.py, testsecondstorad now calls weathercontroller.secondstorads(...) directly, so it tests a pure function. 
Verified that the function’s output exactly matches all three existing test cases (minrad, maxrad, mid_rad).

### DIFF
--- a/smart_control/simulator/weather_controller.py
+++ b/smart_control/simulator/weather_controller.py
@@ -30,6 +30,17 @@ WEATHER_CSV_FILEPATH: Final[str] = os.path.join(
 )
 
 
+def seconds_to_rads(seconds_in_day: int) -> float:
+  """Returns radians corresponding to number of second in the day.
+
+  Args:
+    seconds_in_day: Seconds that have passed so far in the day.
+  """
+  return (seconds_in_day / _SECONDS_IN_A_DAY) * (
+      _MAX_RADIANS - _MIN_RADIANS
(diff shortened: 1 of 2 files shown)


s2t2:
looks like we have some failing tests. please address when you get a chance

Mustaphayinka:
Thanks,looking into the failing tests now.

Mustaphayinka:
Found it, a decorator/function indentation mismatch in weather_controller_test.py that was causing black to fail parsing the file entirely. Fixed and verified with black --check, py_compile, and a direct functional test confirming seconds_to_rads still returns the correct values for all three original test cases.

PR #183, opened 2026-08-22
Interpolate by column name instead of a values property

### LINKED ISSUE #149: Specify Column Name for Replay Weather Interpolation
> In the Replay Weather Controller's  `_get_interpolated_value` function, we are currently passing the list of values to be interpolated, but this requires a separate property for each column of values (e.g. `humidities`, etc.). 
> 
> If possible, let's implement a more flexible interface where we can just pass the column name (instead of the values). 

 _Originally posted by @s2t2 in [#148](https://github.com/google/sbsim/pull/148/changes#r2784670109)_

Mustaphayinka:
Fixes #149

Reworks getinterpolatedvalue so it accepts a column name (str) instead of a pre-fetched pd.Series, then looks up that column internally from weatherdf. This eliminates the need for a separate property for each column, such as temps_f and humidities, which #149 identified as inflexible.
Changes: getinterpolatedvalue now accepts columnname: str rather than values: pd.Series. The unnecessary tempsf and humidities properties have been removed. getcurrenttemp passes 'TempF' directly, matching the column name documented in readweathercsv's docstring, while getcurrenthumidity passes self.humiditycolumn so the existing configurable column-name behavior remains intact.
getcurrenttemp and getcurrenthumidity retain their exact public signatures, so callers elsewhere in the simulator—airhandler.py, simulator.py, solarradiation.py, and others remain unaffected. A search confirmed that none of them access the removed properties or call the private helper directly.
A manual check against the existing test fixture, localweathertestdata.csv, also passed: at 2023-07-01 03:00:01+00:00, getcurrenttemp returns 298.15, matching the value expected by the existing test. For that same timestamp, getcurrent_humidity returns 50.0.

### DIFF
--- a/smart_control/simulator/weather_controller.py
+++ b/smart_control/simulator/weather_controller.py
@@ -30,6 +30,17 @@ WEATHER_CSV_FILEPATH: Final[str] = os.path.join(
 )
 
 
+def seconds_to_rads(seconds_in_day: int) -> float:
+  """Returns radians corresponding to number of second in the day.
+
+  Args:
+    seconds_in_day: Seconds that have passed so far in the day.
+  """
+  return (seconds_in_day / _SECONDS_IN_A_DAY) * (
+      _MAX_RADIANS - _MIN_RADIANS
+  ) + _MIN_RADIANS
+
+
 @gin.configurable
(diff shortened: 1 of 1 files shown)


s2t2:
looks like we have some failing tests

